PR #2, opened 2017-05-17
Static summary

nicoleng12:
- Cleaned up example code from React.js example project
- Added hard coded sample data for patient summary (current staging, pathology results, event dates). The sample data are hard coded strings
- Styled the summary panel


Dtphelan1:
Looks good for now. We can discuss the future of styling sometime this week, but it works for our purposes.

PR #3, opened 2017-05-19
Added Staging Form Component

cjduffett:
Created a new Staging Form component to capture the patient's current cancer stage. The "Prognostic Stage" is automatically computed from the T, N, and M values.

staging-form


Dtphelan1:
Looks good -- I fixed a few small styling things on the data summary section to make sure they look identical.

PR #4, opened 2017-05-23
Selecting summary items

nicoleng12:
Adding interaction between components


Dtphelan1:
This is fine for now; since the goal of this task and this sprint is just to demo something to FW I'm not giving too much thought to if this perfectly captures the react motifs we'll integrate into our code longterm.  Functionality looks great!

PR #5, opened 2017-05-23
Partial solution to inline helper 

Dtphelan1:
Detects a shorthand keyword (.stage) followed by a space. Adds in-line helper text and allows users to edit the newly introduced areas.


gregquinn2001:
can we change .stage to .staging since it seems like staging would be TNM and stage would be roman numeral followed by letter like IIA

Dtphelan1:
Changed terminology @gregquinn2001

gregquinn2001:
lets get rid of work "information" in expansion as well

gregquinn2001:
so we don't have the ability to type in the fields yet? That was part of this task. Do we think we can do it with slate?

Dtphelan1:
Current solution fails to address the continued typing behavior; this code will still be merged in to allow for parallel development on a slate.js drive editor section as opposed to master's current draft.js editor.

PR #6, opened 2017-05-25
Added Dynamic Insertion of Static Data to Editor

cjduffett:
You can now insert structured data into the editor; .NAME, .AGE, .GENDER. This data is passed to the editor as a 'data' prop for now.

### DIFF
--- a/src/ClinicalNotes.jsx
+++ b/src/ClinicalNotes.jsx
@@ -20,41 +20,41 @@ class ClinicalNotes extends Component {
     this.handlePRStatusChange = this.handlePRStatusChange.bind(this);
   }
 
-  handleTextChange(e, val) { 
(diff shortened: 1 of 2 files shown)


gregquinn2001:
Does case matter? I can type ".name" and that will still insert the patient's name?

Also, .name should insert their full name - not just their first I think. Based on properties passed to MyEditor it looks you only did first name

nicoleng12:
This branch successfully meets the task objectives: detect shorthand (.NAME, .AGE, .GENDER), pulls requested data from model based on shortcut, and inserts requested data into the area. Branch successfully updated to handle case insensitive shorthand and enters full name of the patient. We'll figure out the data model later. 

Looks good!

PR #7, opened 2017-05-25
Inline help feature

Dtphelan1:
From task 69 on our board: Allow user to enter shorthand notation resulting in in-line help for sub-field specification to represent structured data concepts defined in SHR.

If user types a number when in one of the appropriate fields: fill in the current field, and move to the next one in sequence.

If user types in 't', 'n', or 'm' when in one of the appropriate fields: switch over to the letter's corresponding  field.

### DIFF
--- a/src/MyEditor.jsx
+++ b/src/MyEditor.jsx
@@ -1,6 +1,6 @@
 // Import React!
 import React from 'react'
-import { Editor, Block, Inline, Raw, Text, Html } from 'slate'
+import { Editor, Block, Inline, Transform , Raw, Text, Html } from 'slate'
 import AutoReplace from 'slate-auto-replace'
 import { List } from 'immutable'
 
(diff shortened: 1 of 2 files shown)


nicoleng12:
Nice! I noticed that if I trigger the staging template (enter the shorthand) after I've already made one, the cursor jumps back to the first instance of the staging template and not the current one that I just made. Is this expected behavior?

gregquinn2001:
@nicoleng12 it's definitely not what we want in the end, but it's one of those bugs he referred to I'm sure in his pull request description. The basic functionality is there well enough to demo maybe. We will use mock-ups from Edwin if they look good or fall back on this initial implementation that Dylan pulled together. Nice job to get it to this point Dylan! It's an important first step and now we can start to refine it after we get some initial feedback tomorrow.

PR #11, opened 2017-05-31
Publish on structured focus

Dtphelan1:
From task 96: Publish events or change state for when editor within staging structured field.

When the staging structured field is initiated in MyEditor, trigger a state change in App that cascades down to the right panel, switching the templates panel to a staging form entry.

Upon blur of focus on the Staging structured field, trigger a state change in App that cascades down to the right panel, switching the current structured field form entry back to templates staging

When focused on the staging structured field in MyEditor, trigger a state change in App that cascades down to the right panel, switching the templates panel to a staging form entry.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -21,47 +21,22 @@ class App extends Component {
       HER2Status: '+',
       ERStatus:   '+',
       PRStatus:   '+',
+      t: "",
+      n: "",
+      m: "",
       SummaryKeyData: '',
-	  WithinStructuredField: null,
-
+  	  StructuredField: null,
     };
-    this.changeHER2Status = this.changeHER2Status.bind(this);
-    this.changeERStatus = this.changeERStatus.bind(this);
-    this.changePRStatus = this.changePRStatus.bind(this);
+    this.changeKeyStatus = this.changeKeyStatus.bind(this);
     this.handleSummaryUpdate = this.handleSummaryUpdate.bind(this);
-    this.handleStructuredFieldEntered = this.handleStructuredFieldEntered.bind(this);
-    this.handleStructuredFieldExited = this.handleStructuredFieldExited.bind(this);
-  }
-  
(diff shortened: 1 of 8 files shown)


gregquinn2001:
Why the rename of WithinStructuredField to StructuredField? I like the old name better as that more indicates it's meaning. It doesn't just hold a structured field but instead references the structured field we are currently within where null means we are not within one right now.

Dtphelan1:
I didn't understand the metaphor of being 'within', because the information is held at the global App level. In that way, I didn't find the naming intuitive. If you feel strongly about having a preposition or adj in front of StructuredField, since it's not the structured field and we might want to disambiguate, we can talk about potential naming candidates or we can revert to the old name. Either way, your call.

gregquinn2001:
Also, why the removal of the structured field entered and exited events in favor of key status event? Again, I'm not sure the name change is intuitive so let's discuss your thinking.

gregquinn2001:
Let's discuss. Call me via skype when you get a few minutes. It's not my call. It's a discussion.

gregquinn2001:
When I click to after the structured field and start typing it never sends an exit event for the structured field.

gregquinn2001:
If I enter the values it successfully sends exit event and then when I left arrow back into the structured field, I get enter event and then when I right arrow out I get exit event correctly so maybe it's only when the structured field hasn't been fully entered that the exit isn't working...

Dtphelan1:
I've made appropriate changes in light of the discussion above. 

As for your newest bug, that sounds like an implementation issue that hopefully we can dodge in the demos. I think capturing it as a jira issue is probably a good idea for now.

gregquinn2001:
sounds good. if you can capture it and move it to the top of backlog to make sure we review it next backlog grooming meeting that would be great

Dtphelan1:
Added bug to the top of the backlog. Feel free to merge if the branch is looking good.

PR #13, opened 2017-06-01
Update staging state from right form changes

nicoleng12:
When user selects t, m, n values from the staging form in the right panel, staging information (t, m, n and prognostic stage) are sent back to the App component to then update the children components with the updated staging information.

Tested manually. In top left menu, select "Enter Staging" and in the Select Staging right panel click on values for tumor size, node, and metastasis. Prognostic stage is updated and when select "Exit Staging" and the re select "Enter Staging" previous input is persisted.  Console window in browser shows that app component is getting the updated staging information.

Worked with Greg to fix updating bug. Greg cleaned up code to break out staging information so each piece of data has its own update function.

Worked with Dylan to fix merging conflict between his clinical notes code and the handling of updated staging information

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -26,7 +26,7 @@ class App extends Component {
             ERStatus: '+',
             PRStatus: '+',
             SummaryKeyData: '',
-            WithinStructuredField: null,
+            withinStructuredField: null,
         };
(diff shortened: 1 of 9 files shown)


Dtphelan1:
Looks good and the functionality checks out on my end.

PR #17, opened 2017-06-06
Height fix

Dtphelan1:
Provides a class to all panel's who establishs a minimum height of 600px. This is an improvment over the current state of heights in the dashboard, but in the future a dynamic sizing solution will be optimal.


gregquinn2001:
on my laptop screen, the summary is a little longer than the other 2 panels.
[Flux Notes Fixed Height Issue Screenshot.docx](https://github.com/standardhealth/flux/files/1055630/Flux.Notes.Fixed.Height.Issue.Screenshot.docx)

Dtphelan1:
The change sets a minimum height, but if the panels overflow beyond that the white will stretch accordingly. Fixed the height to be greater in hopes of avoiding this.

gregquinn2001:
I guess it's ok for now if we can't just set it to 100% or 720 whichever is lower? On my latop screen it now scrolls. Demos usually can't use large screens so I think having it look good on a laptop makes sense.

gregquinn2001:
(I mean demos over skype of course which I think we'll need to do)

nicoleng12:
It looks good on my end on my laptop screen.

Dtphelan1:
1. I think the scrolling is necessary on a laptop no matter what we do for this, since the data summary seems to take up >= 700px no matter what. 
2. The fact that it doesn't look good in some environments will happen no matter what since we're implementing a sub-optimal solution. I can agree that we should build for what we think will make the demo best, and if we think that means making the height 100% or 100vh I'm fine with that. Just let me know what you want to see in order to put the PR through. 

Cheers,
- Dylan

Dtphelan1:
N.B. the scrolling issue will likely disappear if we implement a hidden right panel that we can pull out -- that'll free up plenty of real estate.

gregquinn2001:
ok. I'll merge this as best for now then and we'll wait for that final solution. If we have issues during demos, we can always create a new item to address it

gregquinn2001:
oh I think I may have been zoomed in and that's why it needed scrolling! Sorry. my bad. It's probably good the way you have it for demos.

PR #18, opened 2017-06-07
Remove warnings

Dtphelan1:
Removed code that contributed to warnings during site generation, keeping code in line with best practices.


nicoleng12:
This looks good to me. There are no more warnings in the console. Unused imports are removed as well as old code (code for the checkboxes in the summary panel). In addition Dylan made some syntax fixes so that the code is cleaner. Greg if this looks good to you I will merge it

PR #19, opened 2017-06-07
Color in summary panel

Dtphelan1:
From task 145: Make values in summary panel blue.

I've colorized the values in the data summary, but honestly I don't feel to great about it. A lot of the values are just +/-'s or dates that mean nothing without the thing which it is a value for, and the colorizing of all values in the summary panel makes for such a density of color that it doesn't really draw ones eye to any one thing in particular. I've made a PR for it, and can merge it in if people think it's better than no highlighting, but after seeing it in action I think we might be better off leaving our layout as is, maybe brainstorming instead a different significance for color encoding (e.g. that the value it corresponds to Structured data in the current note, the value corresponds to data pulled from EHR [suggested in Involutions mockups], the value needs clarification [in the case where context was insufficiently clear], etc...)

That said, the code contains tweaks that I do think should go in, even if we remove the classNames responsible for highlighting. Specifically, I've separated the display strings into names and values, moving the names into the list elements themselves and storing the values as variables.

### DIFF
--- a/src/DataSummary.jsx
+++ b/src/DataSummary.jsx
@@ -56,30 +56,30 @@ class DataSummary extends Component {
 
         // Current Staging
 
-        const StageString = `Prognostic Stage: ${this.props.stage}`;
-        const StageSubElementsString = `T: ${this.props.tumorSize}, N: ${this.props.nodeSize}, M: ${this.props.metastasis}`;
-        const showMissingString = this.props.hasStagingData;
-
-        if (showMissingString) {
+        const Stage = `${this.props.stage}`;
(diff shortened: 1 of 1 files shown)


nicoleng12:
I don't feel too strongly either about the coloring. I agree that it doesn't really draw ones eye to any one thing but what it does do well is separate the labels for the values which I think is useful.

I think we should still go ahead and merge it because like you wrote Dylan, there are some code fixes in here for the display strings that should be merged. Unless anyone has strong reservations, I'm going to merge this

Dtphelan1:
If we want the other changes and don't want the coloring, that would only require another 3 minutes of changes. I guess that's the question I'm asking -- but we can wait until the meeting to discuss.

PR #20, opened 2017-06-08
Disease tray header

nicoleng12:
Added section for disease tray header. Modeled it after Involution's mockup. Added buttons to switch between treatment trays.

*NOTE: I am not a fan of how the buttons look or where they are placed. I put the buttons in to at least get a proof of concept but I am definitely open to better styling decisions for this.


gregquinn2001:
agree about the buttons. I was thinking a little more subtly like <> as a button next to the header saying "Lobular carcinoma of the breast" and get rid of "Current Diagnosis" label. What do you think?

gregquinn2001:
better. I like the buttons here are my nits. Let me know if you agree/disagree. I don't like the label "Current Diagnosis" with the arrow buttons to the right of that. We are not cycling through current diagnoses (is that the plural?) only. If we feel we definitely need a label, maybe "Treatment"? My preference would be to just get rid of the label and put the buttons next to the "Lobular carcinoma of the breast".  That also frees more real estate up for the summary data. Make sense? Finally, it would be cool if we disabled the right button maybe because we can explain it defaults to the most current treatment going on? What do you think?

Dtphelan1:
I see what you mean regarding the Diagnosis term. I think that the header might

Dtphelan1:
Take a look now. Let me know what you think.

gregquinn2001:
I like it better. do you think it's unclear what it represents? I'd say keep it for now and let's see what Edwin comes up with maybe?

Dtphelan1:
I don't think it's too unclear, but I think that it looks better with the small heading, but not enough to hold this up. I think we should just merge this, and see what Invo thinks. Good?

gregquinn2001:
Yeah I agree. I said let's keep it for now also. I'll merge it!

PR #22, opened 2017-06-12
Included "Add" button to insert summary info into clinical notes

nicoleng12:
Added plus icon next to the items in the summary panel. Moved onclick event from the label (span text) to the add button. 

The functionality is the same except instead of clicking on summary item text, click on the plus button next to it. Clicking the button inserts text into a span under the clinical notes.

Next steps:
- Insert text into clinical notes (MyEditor)
- Determine what should be displayed in clinical notes

### DIFF
--- a/src/DataSummary.css
+++ b/src/DataSummary.css
@@ -105,6 +105,18 @@
     padding-left: 20px;
 }
 
+#data-summary #summary-condition-details .summary-condition-button .fa {
+    font-size:.8rem!important;
+    color: #00d6b2 !important;
+}
+#data-summary #summary-condition-details .summary-condition-button {
+    height: auto!important;
+    width: auto!important;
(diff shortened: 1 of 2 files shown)


gregquinn2001:
I like it. We'll see what others think!

gregquinn2001:
Dylan, if you're ok with it, please merge it

gregquinn2001:
should we have the + buttons for the patient demographics too? name, age, and gender? maybe that also points to implementation approach as a shortcut for shorthand (clicking on + icon next to name would be like typing .name for example

Dtphelan1:
It looks good to me. I'll merge it as is for now, and if we feel we want the buttons on demo information we can make a separate small task for that. But since the Summary information all has clickable buttons I'm marking this as done.

PR #23, opened 2017-06-13
Toolbar header

Dtphelan1:
From task 108: Add styling toolbar to editor

Styling toolbar is now available and functional. Making a newline is a somewhat temperamental process, but the exact dimensions of each corner case still need to be fully documented. That said, this task in and of itself has been done.


gregquinn2001:
looks good to me. there are a few extra spaces being inserted here and there but it certainly covers what we needed for this jira item. I'm ok to merge.

PR #24, opened 2017-06-13
Added react-calendar-timeline component and data

cjduffett:
I crated and populated a timeline with sample data pulled from the mockups.

You will need to install additional dependencies to run:
```
yarn add react-calendar-timeline
yarn add moment interact.js
```


gregquinn2001:
I suppressed left sidebar to make it look more like concept.

cjduffett:
Ah, good catch. Are you satisfied with the layout at this point?

gregquinn2001:
I think it's a good start and matches a subset of the concept as we discussed. Want to look at getting the key event dates or changing some of the medications to chemotherapy sessions that include cycle and dosage information (the latter would be a lot easier and help it align with what we heard at DFCI)? You can create a subtask or subtasks to cover it/them. If you think that's too difficult, we'll go with it as is.

gregquinn2001:
You can also run it by Andre and see what he thinks based on his shadowing experience.

cjduffett:
I made a slight tweak to the chemo meds: They're typically given in 4-6 cycles, one cycle per month.

Dtphelan1:
I'm assuming this is good to merge? Sounds like it meets enough of the criteria we set out in this task. If we need to, we should make more Jira items for any future styling/features, but we should merge and close when possible.

PR #25, opened 2017-06-15
Added breast cancer staging computations

cjduffett:
Added methods and lookups to convert T,N,M to prognostic stage and prognostic stage to T,N,M for the 5th, 6th, and 7th editions of clinical staging for breast cancer.

```js
const staging = require('lib/staging');

// uses only the 7th edition for these computations:
const prognosticStage = staging.breastCancerPrognosticStage('T1', 'N2', 'M0');

// 5th, 6th, and 7th editions are all valid (defaults to 7th):
const values7thEd = staging.breastCancerPossibleTNM('IIB');
const values6thEd = staging.breastCancerPossibleTNM('I', 6);
```

The `prognosticStage` is returned as a display-ready string, e.g. `'IIIA'`.

Possible TNM values are returned as 2D array, e.g.:
```js
[ ['T1','N1','M0'], ['T1','N2','M0'] ]
```

Also added unit tests for these library functions. I'm not sure how `yarn test` is configured for react, so I've just been using good ol' fashioned `mocha`:

From the top-level project directory, simply run:
```
mocha
```

You should see something like:
```

  breastCancerPrognosticStage
    ✓ should compute valid prognostic stages given T, N, M
    ✓ should return null for invalid T, N, M values
    ✓ should return null for undefined staging given T, N, M
    ✓ should be case-insensitive

  breastCancerPossibleTNM
    ✓ should return a list of possible T, N, M tuples given a prognostic stage
    ✓ should return all permutations of T and N for stage IV
    ✓ should return an empty list for an invalid staging edition
    ✓ should return an empty list for an unknown prognostic stage
    ✓ should default to the 7th edition if none is specified
    ✓ should be case-insensitive


  10 passing (12ms)
```

Now that we have unit tests for something, perhaps we should also setup Travis CI?

### DIFF
--- a/lib/staging.js
+++ b/lib/staging.js
@@ -1,0 +1,63 @@
+const lookup = require('./staging_lookup');
+
+// Computes the prognostic stage of breast cancer given T, N, M.
+// Uses the 7th staging edition, see:
+// https://cancerstaging.org/references-tools/quickreferences/Documents/BreastMedium.pdf
+exports.breastCancerPrognosticStage = (t, n, m) => {
+  // Regardless of edition, metastatic cancer is always stage IV
+  if (m.toString().toUpperCase() == 'M1') {
+    return 'IV';
+  }
+
+  // 7th edition staging, lookup[T][N]
+  // null return values mean the staging is undefined for those T,N,M values
+  const ti = (lookup.getTsForEdition(7)).indexOf(t.toString().toUpperCase());
+  const ni = (lookup.getNsForEdition(7)).indexOf(n.toString().toUpperCase());
+
+  if (ti === -1 || ni === -1) {
+    // Urecognized T or N value
+    return null;
+  }
+  return lookup.getTableForEdition(7)[ti][ni];
+};
+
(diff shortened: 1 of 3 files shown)


gregquinn2001:
This looks great Carlton! Can we integrate this into Flux Notes? I guess for now that would just mean replace our function that calculate stage from TNM with a call to your library?

gregquinn2001:
I'm fine with merging this. Not sure if Nicole or Dylan want to take a look

cjduffett:
I want to make one minor adjustment to see if I can get it to work easily with `yarn`. That way we're using `yarn` for our full build/test/run process. Let's review at scrum before merging.

cjduffett:
Shelving the testing for now. Ready to merge.

Dtphelan1:
Looks good on my end. I'll merge it and Carlton said he'll make a Jira task for integrating this into the current infrastructure. It'll go in the backlog and we can revisit it next sprint planning meeting.

PR #26, opened 2017-06-15
Insert summary data into clinical notes

nicoleng12:
Clicking on the buttons inserts the summary data into the clinical notes at the last cursor location. Click somewhere in clinical notes and then click the add button to insert the text at that location. 

Renamed some functions and property names so that it the functionality is more clear (related to inserting text)

Questions/looking for feedback:

- Right now the text that is inserted mirrors what is displayed in the summary panel. What do we want the text to be? I grouped sub items with main item and got rid of redundant button for summary sub items (i.e prognostic stage, Diagnosis, and Surgery)

-  In MyEditor.jsx, inserting of the text is triggered by componentDidUpdate. This function is called when the state is changed. The if statement for setState prevents componentDidUpdate from entering an infinite loop. In componentDidUpdate, handleSummaryUpdate is called which inserts the text. The text that is inserted is the property value because that is the most up to date. I had tried to insert the state value but for some reason when 'this.state.itemToBeInserted' is set to 'this.props.itemToBeInserted', the value is not set right away and is one value behind. You can comment out line 209 and replace it with 208 to see this in action


gregquinn2001:
So in the future, we will need a way to define a shortcut via a generic interface and part of that will be what text including structured data and fixed text gets inserted. I think we can view the clicks on summary items as the equivalent of typing a shortcut into the notes editor. Clicking diagnosis date insert button for example, is equivalent to typing .diagnosis date in the editor. that structured field should be associated with the shortcut (obviously. For now though, I think we should make them insert the simple data value that each field represents. Clicking on diagnosis date should insert just the date itself for example.

gregquinn2001:
It's inserting at the beginning right now. Would it be easy to make it insert at the end at least?

nicoleng12:
It inserts where the cursor is at. So if you start the app and haven't clicked in clinical notes yet it will default the cursor to the beginning I think. I'll look into seeing how hard it is to put it at the end. Dylan, since you've worked with slate js in depth, do you know if this is difficult to do as opposed to inserting at the cursor?

I will change the inserted text to be just the value and not include the label

Dtphelan1:
In the future we should work on that generic interface, I agree with Greg. If we come up with a good implementation of that I think that solves your the first problem you mentioned: "Right now the text that is inserted mirrors what is displayed in the summary panel. What do we want the text to be? I grouped sub items with main item and got rid of redundant button for summary sub items (i.e prognostic stage, Diagnosis, and Surgery)" 

As for the second issue of infinite loops, I was able to achieve the same behavior we wanted by using a different lifecycle function "componentWillReceiveProps". This triggers anytime a change has been mad eto the properties being passed to the component in question. By using this function, we circumvent the need to track the previous thing inserted by storing it in our state. 

Moving forward, we'll probably want to use a different mechanism for this; for example, we could look at a buffer-like implementation where summary panel loads a buffer to be pasted and clinical notes empties the buffer after pasting, removing the need to track previous pastes and enabling users to paste the same structured data multiple times in a row (for instance, in different locations on the page).

Let me know if this works, and if so I think we're good to merge.

nicoleng12:
Thanks Dylan! "componentWillReceiveProps" is exactly what I was looking for and somehow missed and used "componentWillUpdate" instead and ended up doing weird code work arounds for the infinite loop

PR #27, opened 2017-06-15
Integrating staging library functions

cjduffett:
Added the new staging library for computing prognostic stage. Changed the tumorSize, nodeSize, and metastasis variables to be strings representing the T,N,M values.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -18,11 +18,9 @@ class App extends Component {
     constructor(props) {
         super(props);
         this.state = {
-            hasStagingData: false,
-            prognosticState: 0,
-            tumorSize: 0,
-            nodeSize: 0,
-            metastasis: 0,
+            tumorSize: '',
+            nodeSize: '',
+            metastasis: '',
             HER2Status: '+',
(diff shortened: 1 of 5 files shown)


Dtphelan1:
It seems like we've re-coupled the staging values (T,N,M) with the Stage value, something we'll have to undo when we begin moving towards new in-line helpers like' .stage'. Since that won't be necessary for the demo, I'm fine merging this as is.

Dtphelan1:
Looks like the staging library makes an assumption wherein if the M value isn't M1, it's assumed to be M0. This results in staging information before the M value has been entered and is incorrect. Once this is addressed it looks ready to merge.

cjduffett:
Fixed with https://github.com/standardhealth/flux/pull/28

Dtphelan1:
So it looks like a few value combinations are undefined. Is this expected behavior as far as the manuals are concerned? If so, it looks clean to merge

Dtphelan1:
Was able to confirm that some of these combinations are undefined in the AJCC manual. I'm good to merge.

PR #28, opened 2017-06-15
Added check for invalid M values

cjduffett:
Fixes https://jira.mitre.org/browse/ASCODCP-193


gregquinn2001:
this works for now so let's merge. We'll have to revisit once our concept is confirmed and as we add additional cancer types and/or edition support

PR #30, opened 2017-06-19
Autocomplete v1

Dtphelan1:
From task 73: Autocomplete shorthand names during notes entry for staging shorthand initially

First stab at autocomplete includes: 
- load possible matches from structuredDataRaw.js
- dropdown of potential matches, 
- update of matches on typing, 
- update of matches on deletion, 
- change current match via hover,
- change current match via arrowkeys (up and left move up; down and right move down),
- insertion of match via enter, 
- insertion of match via click, 
- cancel autocomplete via spacebar, 
- auto-focus on "firstSelection" as specified in structuredDataRaw for .staging,
- use of 'positions' library to find location for dropdown

Additional features include: 
- From task 140: Staging inline helper is inserted onto a new line; insert onto the same line.


Dtphelan1:
Also: I've removed the TMN toggling. We can put it back if we want, but I didn't see it sticking around and figured i'd knock it out while I was at it.

nicoleng12:
Tested all the features above and it works as expected. Just noticed a couple of things:

- Is it expected behavior that when users type ".stag", cancel the autocomplete by typing space, then backspace and try to continue typing that the autocomplete is still cancelled?

- When inserting the inline helper for staging via space after ".staging" or pressing enter for the autocomplete, the focus is set to the first value, but when clicking "staging" in autocomplete, no focus is set

- When delete everything in the clinical notes and insert inline helper, no focus is set whether inserting the helper via autocomplete (enter button and mouse click) or pressing space

Dtphelan1:
In order: 
- Yes, since the only way to enter the autocomplete state right now is by entering a '.', and entering a space exits the current mode. This should be avoided in the demo
- This is not expected. I'll look at this. 
- This is a known issue not only with the dropdown but also with the regular inline helper.

gregquinn2001:
Maybe after resolving 2nd bullet in Nicole's comment that you say you'll into, we can merge this?

Dtphelan1:
Doesn't look like it's trivial to resolve. I was looking at this yesterday, and we might just want to make that a larger task in the next sprint. It looks like the editor is losing focus when I click on something else. Preventing the event default behavior doesn't fix this either.

gregquinn2001:
ok. should we create a jira bug against it at top of backlog and then we can merge it as is?

Dtphelan1:
Jira bug has been logged. Feel free to merge if it's good!

gregquinn2001:
I guess I should verify that the other 2 issues Nicole observed are also logged?

Dtphelan1:
They're logged aswell

PR #31, opened 2017-06-19
Redesigned Summary Panel and Styling Refactoring

cjduffett:
In a word, I redesigned the Data Summary panel and heavily consolidated some confusing styling practices. I also broke up our component src/ directory into several subdirectories for related components.

**I unified and consolidated inconsistent app styling:**

1. Any reusable app-level components (e.g. the panels) are now styled in one place
2. I simplified the styling of the staging form (also added borders to make the buttons pop more)
3. I updated the positioning/styling of the panels to maximize the use of screen space

**I broke the data summary panel down into several components:**

1. DataSummaryPanel - the top level UI panel
2. SummaryHeader - the picture of Deb, and her basic info
3. ConditionSelection - tool to switch between different conditions using arrows
4. DataSummaryTable - table used 3 times to show Diagnosis, Pathology, and Genetics

**I moved most of the data in the app up to the application level:**

1. Most data displayed in the app is now kept and modified only at the App.jsx level

Of course, I made sure the existing data bindings still work:

1. Entering T,N,M from clinical notes or staging form
2. Clicking on a captured structure data element to insert it into clinical notes


cjduffett:
Also fixes https://jira.mitre.org/browse/ASCODCP-202

Dtphelan1:
Made some small improvements to the scrollbars styling. There are some other things I want to say, but they can wait until scrum.

nicoleng12:
Does this branch have the latest version of master merged into it? Updating staging information from the right panel does not update the summary form or the clinical notes. Updating staging information does update the right panel but does not update the summary panel

nicoleng12:
Nice styling! Nice job making the summary panel look like the mockups from Edwin. The functionality still works (the bindings across all three components work for updating staging data). Maybe we can make the font size in the summary panel for the data elements a little bit larger or bolder? They are slightly difficult to read. User cannot enter Tis or N1mi in the inline helper in clinical notes but that issue is covered in jira task 194

nicoleng12:
Nice. I like the buttons that appear on hover

gregquinn2001:
Was font size updated? It looks better to me than first time I saw it. I'm still thinking we need a UI that makes it obvious without hovering that you can interact with data items. Not sure it will be obvious from + what button does exactly, but at least they will see the button and know they can do something.

For demos, I use the diagnosis date so can we put dates back into the summary even though they are moved in the mock-ups to their own tab since we don't have the tabs yet?

gregquinn2001:
Clicking arrows in summary should not change condition to hypertension unless we are going to have a complete summary for that which we won't so let's just put it back to the button doing nothing

gregquinn2001:
Need to add a slash character between M value and the stage. e.g.:

T2 | N0 | M0 / Stage IIA

cjduffett:
Thank you for the feedback @gregquinn2001, I made your requested updates.

gregquinn2001:
Still not obvious to me when looking at summary panel that you can interact with things on it. It appears to be view only. The plus buttons (for now at least) should always be visible.

gregquinn2001:
Also the inserts have been changed from what Nicole had. The inserts should only insert the data - no label.

cjduffett:
@gregquinn2001 - your additional changes are now incorporated.

PR #32, opened 2017-06-21
Right panel hover on buttons

nicoleng12:
Added tool tips to the buttons in the right panel so that when users hover over the buttons, additional information about that value is displayed.

- Added descriptions (text displayed in tool tips) to the T map, N map, and M map. Updated all editions (5th, 6th, and 7th) based on cancer staging manual pdfs referenced in staging_lookup.js

- Updated variables in StagingForm.jsx to make sure data bindings still work (updating staging information)

Please note:

- For some of the buttons, the tool tip text is VERY long (i.e T4, N1, N2, N3, M1). For the text, I copied and pasted the information from the breast cancer pdf. I don't know how much of the text is actually needed by clinicians. There are some options for handling the long text: cut down on the amount of text displayed (need to make sure we aren't removing anything essential) or instead of a tool tip make a help/info icon that when clicked opens a modal dialogue that displays all of the text

- I am open to other styling ideas and suggestions for the tool tips

### DIFF
--- a/lib/staging.js
+++ b/lib/staging.js
@@ -17,8 +17,8 @@ exports.breastCancerPrognosticStage = (t, n, m) => {
 
   // 7th edition staging, lookup[T][N]
   // null return values mean the staging is undefined for those T,N,M values
-  const ti = (lookup.getTsForEdition(7)).indexOf(t.toString().toUpperCase());
-  const ni = (lookup.getNsForEdition(7)).indexOf(n.toString().toUpperCase());
+  const ti = (lookup.getTsNamesForEdition(7)).indexOf(t.toString().toUpperCase());
+  const ni = (lookup.getNsNamesForEdition(7)).indexOf(n.toString().toUpperCase());
 
   if (ti === -1 || ni === -1) {
     // Urecognized T or N value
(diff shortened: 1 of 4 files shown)


Dtphelan1:
Nicole this isn't generating on my end. I'm getting console errors
hover-errors

Dtphelan1:
This looks really good. Based on the conversation we just had offline, it sounds like we might want to add ~2 classes for tooltip sizing, so we can have wider tooltips for larger blocks of text without making the small tooltips look silly sitting in a really wide area. 

Other than that I think this looks pretty great.

gregquinn2001:
Looks great except for the super long one (N3) which goes off my screen at the top. Is there a way to make sure it fits on the screen?
Also, I think the prognostic stage calculation is broken? Only M1 gets a value to calculate otherwise I seem to get undefined - even for T2 N0 M0

Dtphelan1:
Good catch Greg -- it looks like calculations are broken on my end too.

nicoleng12:
Nice catch. I found where it's broken so I'm fixing that now. I'll also add functionality to toggle between different width size tool tips. Because the buttons are generated in a loop, instead of making 2 classes for small width and large width (not sure how to set different classes if it's generated in a loop), I'm adding ids to the span text and will set the size using the id. If there other ideas for this I'm open to suggestions.

nicoleng12:
Ok I addressed both issues. When you get a chance, please review again and let me know if you find anything else. Thanks

PR #35, opened 2017-06-27
Shorthand dropdown and @ for suggestions

Dtphelan1:
Code should be refactored into a series of  at least two components: 
1. A generic 'portal' component that corresponds to the dropdown. 
2. An autocomplete component that uses the portal component. 

We can then instantiate two of the autocomplete components, one for shorthand and one for the inline helpers.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -48,16 +48,16 @@ h3 {
 }
 
 
-::-webkit-scrollbar {
+.App-content ::-webkit-scrollbar {
     width: 12px;
 }
 
-::-webkit-scrollbar-track {
+.App-content ::-webkit-scrollbar-track {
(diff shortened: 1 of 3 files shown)


nicoleng12:
The @ for suggestions works as expected and the dropdown also works as expected. Selecting an option from the drop down via typing and pressing space, via enter for selection, and via mouse click works. 

I noticed that on mouseover for the dropdown there is an error in the console that "updateCurrentShorthandMatch" is not a function. There's a typo in MyEditor.jsx on line 692. "updateCurrentShorthandeMatch" should be "updateCurrentShorthandMatch"

I fixed it and pushed it to your branch

Dtphelan1:
Thanks for the catch. 
If there isn't anything else coming to mind @gregquinn2001  or @cjduffett I'm going to push it. 

- Dylan

PR #37, opened 2017-06-29
View existing notes

nicoleng12:
Added tab component to the left panel. Leveraged material ui's tab component and overwrote material ui's styling. Modeled the tab view after Involution's mockups. User can click on the tabs to toggle between problem summary and clinical notes.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -18,7 +18,7 @@
 }
 
 .trio {
-    min-height: 750px !important;
+    min-height: 810px !important;
 }
 
 .panel-content {
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -36,7 +36,24 @@ class App extends Component {
                 address: {
                     city: "Boston",
                     state: "MA"
-                }
+                },
+                previousNotes: [
+                    {
+                        date: "02.AUG.2015",
+                        subject: "Clinical follow-up",
+                        hospital: "Dana Farber Cancer Institute",
+                        clinician: "Dr. Smith"
+                    },
+                    {
+                        date: "12.JUL.2012",
+                        subject: "Consult",
+                        hospital: "Dana Farber Cancer Institute",
+                        clinician: "Dr. Zheng"
+                    }
+                ]
+
+
+
             },
             conditions: [
                 {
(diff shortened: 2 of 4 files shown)


gregquinn2001:
looks good and matches concept. I do question whether the list of existing notes should be under the current diagnosis? Does that mean that only notes associated with the condition/diagnosis selected from the drop-down above the tabs will be shown? Not sure how doable that is with data from EHRs so could be interesting...

gregquinn2001:
It's fine with me to merge it!

nicoleng12:
I believe Dylan is looking at it right now as well, so I'll wait for Dylan's feedback then will merge if all is fine

gregquinn2001:
The summary is as wide as the notes area! Why do we have so much extra space on each side of our fields? Can't our summary start on left edge of screen with only a small amount of space as a margin?

gregquinn2001:
"fields" should have been "panels"

Dtphelan1:
1. The width of the summary is necessary as is because of issues we had scrunching the summarys into a small amount of space. At least I think. @cjduffett could verify.
2. The extra space comes out of the max-width restrictions we put on the app, in order to prevent everything from stretching out too much. This looks weird in the current layout. 
3. All these things begin to disappear if we consider the removal of the current permanent right panel and free up that real estate. But a) this is a longer design conversation that relates to expected user experience and b) seems subsidiary when compared to getting basic functionality of new features in, right? 

</2-cents>

Dtphelan1:
Related to the PR: I think it looks pretty good. I want it to look more like the mockups, but that would require a concerted redesign of a few basic things all over the place. So for now I think this is good and alright to merge.

gregquinn2001:
can we just left justify the whole thing?

Dtphelan1:
But then wouldn't that just throw all the margin width on the right?

Dtphelan1:
Either way -- to be clear these issues are separate from Nicole's changes. Assuming they are, I'm making Jira items for your issues. We should consider figuring out a preferred forum for having these general discussions.

Dtphelan1:
Also, just left justifying everything is likely easier said than done, since we've been usinng libraries to integrate the column spacing and sizing of bootstrap into our react project. We might have to refactor a fair amoutn, but I'm unsure.

nicoleng12:
We can use slack to capture these thoughts

PR #38, opened 2017-06-30
Added routing for full and slim UI versions

cjduffett:
I used [react-router v4](https://reacttraining.com/react-router/web/guides/philosophy) to provide client-side routing for a "slim" version of the UI (just templates and editor) or the "full" version of the UI. These are served at the following routes:

`localhost:3000/` (redirects to `localhost:3000/#/`) - serves full UI  

`localhost:3000/#/slim` - serves the slim version of the UI

At present the routing scheme used is called [Hash Routing](https://reacttraining.com/react-router/web/api/HashRouter) and can be used entirely client side. A more familiar version of routing called [Browser Routing](https://reacttraining.com/react-router/web/api/BrowserRouter) is also possible, but requires a backend server to support the views served at each route. For now, hash routing was simplest.

For now the only way to navigate to the slim version of the UI is by modifying the URL in the browser's navigation bar. If you want to click on a button on the navbar to change routes, you'll need to move the `NavBar` component up into the new `App.jsx` file and add [Link](https://reacttraining.com/react-router/web/api/Link) components to switch between routes.

This will be my last PR for awhile. Until then...
Ron Burgundy


gregquinn2001:
Can you switch it so the we have /patient which gives the view where it has a patient context (the current UI) and then the root URL will show the no patient view? If that's a big deal then leave it as you have it and add a jira to change it

cjduffett:
@gregquinn2001 - I made the change. :)

gregquinn2001:
thanks!!

nicoleng12:
This works as expected on my end. URL correctly navigates to slim view and full view (when adding /patient). Editor and staging form still works in the slim view

nicoleng12:
If you still have time, can you add the new technologies that you used for routing to our wiki page?

gregquinn2001:
I think he left Nicole. I'll try to add it

gregquinn2001:
done. I added react-router v4 to list on wiki

PR #39, opened 2017-06-30
remove note header from editor panel when in no-patient mode

gregquinn2001:
pass patient to clinical editor component and use as trigger for whether to include patient header or not.


nicoleng12:
This looks good to me. Toggling between patient and no patient view removes the patient header

PR #40, opened 2017-06-30
Progression shortcut

Dtphelan1:
They said it couldn't be done....

From task 240: Shortcut for #progression in clinical note.

The task has been completed sufficiently to demo. Specifically, typing: "#progression[Stable imaging, pathology]" will produce the text "#progression[Stable based on  imaging, pathology]" and produce a new progression event (reflected by insertion into timeline and logs to console).

Known bugs: 
- For whatever reason the timeline events get a new start date everytime text gets updated while entering progression information. I can remove this if possible, or someone else can by commenting out the componentWillReceiveProps line in the timeline component. 
- The behavior for typing beyond more flexible input is unpredictable right now. 
- Backspacing while typing text can lead to bugs. 
- Selecting off and back into the text while entering progression can lead to bugs. 
- Trying to delete from the last character for a structured element leads to bugs. 

Let me know if this meets a sufficient number of the features needed to merge.


Dtphelan1:
Also: N.b. all inline helpers now use the # with square brackets format. specifically: 
- # marks the beginning of autocomplete options
- [ triggers insertion now, instead of a 'space' 
- ] triggers completion, which is a new feature we hadn't needed until now.

nicoleng12:
When I type #progression[Stable imaging, pathology] in the clinical notes section and press space or enter, I do not see anything being inserted into the timeline... am I missing a step? 

As noted in your list of bugs, it is a little strange to see the timeline events changing position in the timeline as I type progression information. I think we should hide the update

gregquinn2001:
I delete #staging in default note text then type #progression[Stable imaging,pathology] and the new progression item appears on the timeline right by now. However, I can't get "based on" to appear at all.

Dtphelan1:
Did you type a space before the sqaure bracket? You shouldn't

Dtphelan1:
If someone could hide the update that would be fine by me. I might not have access to my machine for some time.

gregquinn2001:
sorry I tried with a space and without a space. I typed exactly what I put above

Dtphelan1:
The based on text isn't displaying? If that's the case then it shouldn't be merged until I can fix this.

gregquinn2001:
exactly. I can't get it to at least. No problem. It can wait until next Wed! Enjoy your weekend and holiday!

nicoleng12:
Yes, the based on text isn't displaying for me. As Greg had written, we can discuss on Wed when we're all back.

Dtphelan1:
I'll have to see someone's machine demo it because it's showing up on my end still... odd....

Dtphelan1:
Three things fixed: 
1. #progression typed out should trigger the autocomplete jump.
2. Typing 'stable' should trigger a jump to the second field, regardless of case sensitivity.
3. Right bracket should pull you to the end of the text block if typed into the reasons structured block.

nicoleng12:
Tested latest fixes and it behaves as expected. Once the updating of the timeline is fixed I think this looks good to merge since it covers what was outlined in the jira task. 

It's not super intuitive that we have to type right bracket to end the text block since an end bracket is already added when the inline helper is displayed. Can we maybe use "enter" instead? Or when the inline helper is inserted, remove the right bracket so that users have to type it? Not really sure what the best approach is. We can probably make a different  jira issue for this

gregquinn2001:
looks like a good start! There certainly are a few usability things to think through as @nicoleng12 mentions but I think this meets the requirements of the jira item and we should capture additional changes via new jira items to be considered during today's backlog grooming. I'm not too worried about the progression items getting added to the timeline right now either if we want to just add a jira bug for that and move on.

Dtphelan1:
I've disabled the timeline feature. If you think this is a sufficient amount to have done for this task, we can merge and define new tasks at the backlog grooming.

nicoleng12:
This looks good to merge, thanks

PR #41, opened 2017-07-03
Progression in summary panel

nicoleng12:
Added progression to the summary panel using the existing data summary table. The summary panel just shows the most recent progression date. If the date is older than 6 months from today's date, then it is displayed as missing progression data. Updated data model to include "reason" field for progression


gregquinn2001:
Not sure how to test the missing scenario but the regular functionality looks like it works. Can we break up the progression and reason into 2 fields under current progression and display the date for the pair. A plus should be able to insert progression value and a separate plus should be able to insert the reason. The other option would be a plus to insert "<progression> based on <reason list>"...

nicoleng12:
To test the missing scenario, you can modify the data model in FullApp.jsx. In the progression section, if you update the most recent entry (Progressing Disease, 2017-05-15) to a date longer than 6 months ago for example "Progressing Disease, 2016-05-15" and then re run the app, you'll see that the box turns red to indicate that progression data is missing.

For the view of the progression information in the summary panel, I modeled it after surgery in the "Procedures" section which also has a date and more detailed information. I can update the data summary table component to break up progression and reason into 2 fields but in terms of consistency for the rest of the summary panel, for other entries that have dates (key dates and procedures) do we also want to follow that model? Make the label the date and the additional information be the values?

gregquinn2001:
yeah I meant to test it without modifying code. No problem. Our default data though probably should show current progression as a missing value

I don't think current progression should be treated like key dates. It's the current progression value with an associated "based on" value which happens to has an "as of" date. In other words, I think the date is secondary to the progression value and associated reason

nicoleng12:
I will edit the summary panel to display progression information as we just discussed in person

Dtphelan1:
Resolved conflicts produced by my data model. Introduced the concept of ID's to act as a temporary mechanism of identity b/t two progression objects.

nicoleng12:
Merged master into this branch and finished editing the view. Please review and let me know if you have any additional feedback

gregquinn2001:
looks good. I'd say merge! I'm assuming the missing case works with the changes too? The header would just be "Current Progression" without an as of date at all and then each name/value would be missing.

nicoleng12:
Oops forgot to remove the date in the header when data is missing. Let me do that now. I'll let you know when that's done and I'll also post a screen shot of what the panel looks like when progression is missing

nicoleng12:
Ok fixed the header. This is what it looks like when progression data is missing: 

image

gregquinn2001:
looks good! merge!

PR #42, opened 2017-07-04
right panel now shows shortcuts when no patient

gregquinn2001:
stays the same when patient

### DIFF
--- a/src/forms/FormTray.jsx
+++ b/src/forms/FormTray.jsx
@@ -16,7 +16,9 @@ class FormTray extends Component {
     let panelContent = null;
     if (this.props.withinStructuredField == null) {
         panelContent = (
-            <TemplateForm />
+            <TemplateForm 
+				patient={this.props.patient}
+			/>
         );
     } else {
         panelContent = (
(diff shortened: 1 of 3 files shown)


nicoleng12:
Switching between patient and no patient view toggles what "templates" or "shortcuts" are displayed. This looks good to me to merge

Dtphelan1:
Minor concern: The header is for templates. Is that what we are providing? Or are we providing the inline-helper shortcuts? If they're templates, I'm good to merge; if not, we should change this header.

gregquinn2001:
Yeah Nicole and I discussed this as well. We're not sure yet if it will be shortcuts and templates (where shortcuts can be a simple template) or just shortcuts. Didn't know what to label that panel. Obviously right now it shows only shortcuts for structured data items so maybe we should just re-label it with "Data Items" or "Structured Data Items" although I've been thinking we'd avoid the term "structured data" in the UI. We can't use "Summary Item" because we have no summary in no patient mode... So, I was a little at a loss for what to name it and whether it would include templates still so I just left it as is...

gregquinn2001:
Should I just change the label to "Data Items" for now and merge?

Dtphelan1:
Yeah that works for now!

Dtphelan1:
After that I'd be happy to merge

PR #43, opened 2017-07-06
support selection event propagation from editor to right panel

gregquinn2001:
selecting one or more characters of text in patient mode causes the right panel to change to some h1 text showing the selected text. that h1 should be changed to be the react component for the curator right panel being done in 248


Dtphelan1:
Works on my end. Going to merge

PR #45, opened 2017-07-06
Right panel data element capture form

nicoleng12:
- In patient mode, when user highlights/selects text in clinical notes, the data capture form is displayed

- User selects what type of data is missing from the drop down. Currently selecting progression and toxicity does not bring up the point of sales displays since they have not yet been created. 

- Selecting staging as the missing data brings up the staging form within the data capture panel

- Staging form within the data capture panel works and updates staging information in the summary panel and the inline helper

- Updating the staging information in the notes also updates the staging form in the data capture panel

### DIFF
--- a/src/forms/DataCaptureForm.css
+++ b/src/forms/DataCaptureForm.css
@@ -1,0 +1,5 @@
+.btn_template {
+  float: left;
+  clear: left;
+  margin-bottom: 20px;
+}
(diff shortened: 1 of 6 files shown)


gregquinn2001:
looks good to me. I'll create a jira to work on the interaction between the curation panel and the notes. we seem to lose the view of the selection right now when focus goes into the right panel. I think we can merge.

gregquinn2001:
jira 286 created

PR #46, opened 2017-07-06
Added copy button to editor tool bar

nicoleng12:
Added copy button to editor toolbar.

- Copy button only shows up when in no patient mode

- Added cursor change to pointer when user hovers over copy icon (consistent with the other buttons in the tool bar)

- Hovering over copy button changes copy icon from grey to black 

- Clicking the copy button currently outputs a console log ("clicked copy")

### DIFF
--- a/src/notes/ClinicalNotes.jsx
+++ b/src/notes/ClinicalNotes.jsx
@@ -131,6 +131,7 @@ class ClinicalNotes extends Component {
             metastasis={this.props.metastasis}
             data={{patient: {name: 'Debra Hernandez672', age: '51', gender: 'female'}}}
             itemToBeInserted={this.props.itemToBeInserted}
+            patient={this.props.patient}
 
           />
         </Paper>
--- a/src/notes/EditorToolbar.css
+++ b/src/notes/EditorToolbar.css
@@ -57,3 +57,11 @@
   margin: 10px 0;
   width: 100%;  
 }
+
+#copy-button a {
+    color: #ccc;
+}
+
+#copy-button a:hover {
+    color: black;
+}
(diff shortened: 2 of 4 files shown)


nicoleng12:
The copy button is subtle. Here's a screenshot:
image

gregquinn2001:
It's definitely there. We should add tool tips to all those buttons eventually. I'm going to merge it

nicoleng12:
Thanks!

gregquinn2001:
The other issue is that a classic copy button in a toolbar would copy the current selection I believe. This behavior should be different so we may want a different button (maybe right justified and just the words "Copy All"???? Before I merge, what do you think?

nicoleng12:
I originally had a button in there and thought it looked a bit off with the other buttons as icon buttons. I can always add a tool tip indicating it's a copy all?

gregquinn2001:
I think that icon is typically associated with the normal copy function which copies the contents of the selection to the clipboard. That's my concern. In addition, this button is a little different; either it should clear the text entry field or we should have a clear button or we should get rid of the text editor altogether as we discussed in backlog grooming!

Dtphelan1:
what's this clear functionality you are thinking about? Also, I can see the confusion from selection copy vs copy all... We can maybe put the button somewhere in the header, with the details of the note?

Dtphelan1:
Or just have it copy the selection, and when there is no selection copy everything?

gregquinn2001:
I think the "copy" button functionality is different than the traditional text editor toolbar copy button is what I was trying to say. This button is more of a "save" or "done" button for when they've created their shortcut in the correct format and then want to put it into their EHR clinical notes editor.  We don't want the user to have to select everything or deselect everything in order to copy all content to the clipboard when they are done. There's also the issue of whether we should have a text editor at all in no patient mode.... We're hoping to discuss after scrum today.

Dtphelan1:
Good with me!

nicoleng12:
Based on our discussion after scrum, I'm going to merge this

PR #47, opened 2017-07-06
Use placeholder

gregquinn2001:
changed the editor to use a placeholder (different text depending on patient or no-patient mode) instead of fixed text to start

### DIFF
--- a/src/forms/TemplateForm.jsx
+++ b/src/forms/TemplateForm.jsx
@@ -10,7 +10,7 @@ class TemplateForm extends Component {
   constructor(props) {
         super(props);
 
-		if (this.props.patient == null) {
+		if (!this.props.patient) {
 			this.state = {
 			  heading : "Data Items",
 			  templates: ['progression', 'toxicity'],
--- a/src/notes/ClinicalNotes.jsx
+++ b/src/notes/ClinicalNotes.jsx
@@ -131,7 +131,7 @@ class ClinicalNotes extends Component {
             metastasis={this.props.metastasis}
             data={{patient: {name: 'Debra Hernandez672', age: '51', gender: 'female'}}}
             itemToBeInserted={this.props.itemToBeInserted}
-
+			patient={this.props.patient}
           />
         </Paper>
       </div>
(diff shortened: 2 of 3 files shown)


Dtphelan1:
Looks good on the whole. As a general rule of thumb, let's try to use [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) checks as we type up our code, otherwise Javascript will do it's earnest to come up with funky behavior.

gregquinn2001:
I actually originally had === for that check but then it doesn't work. I always get else clause in both no patient and patient mode. Try it if you want. I changed it to == and it works. Any idea why?

Dtphelan1:
Hm... odd. 
Something worth looking into. I think it's fine now if that's the case

nicoleng12:
When I run it, I keep getting the if case for both patient mode and non patient mode...this is when using ==

I noticed that In other places in the code we also have null checks using double equals instead of triple.  Instead of I checking for null, we can always just check it it exists so instead of `if (this.props.patient == null)`  we can just do `if (!this.props.patient)`

gregquinn2001:
wow that's strange! I'll try the exists check as you suggest @nicoleng12

gregquinn2001:
this is strange. Now it's not working with any of the 3 cases!

gregquinn2001:
when I rebased to master I lost passing in patient (thanks @nicoleng12!). I'll fix and commit

gregquinn2001:
ok. all changed now to use the existence check (well non-existence check with the !)

nicoleng12:
ok this looks good on my end now

gregquinn2001:
I'll merge and create a JIRA to replace existence and == and === check with lodash equivalents? It will be in backlog at least that we have these in the code that should be changed. Make sense? Ok to merge @Dtphelan1 ?

gregquinn2001:
wait, how does lodash help us with checking if a property exists or not?

gregquinn2001:
would we use _.isUndefined to check for the existence of a property?

nicoleng12:
Yes I think so. You can also check if null: https://lodash.com/docs/4.17.4#isNull

gregquinn2001:
created jira. I'm merging

PR #48, opened 2017-07-07
Added hover over for insert buttons in summary panel 

nicoleng12:
When the user hovers over the insert buttons in the summary panel, the cursor changes to a pointer (similar behavior to the buttons). Also added cursor change to arrows in condition selection section of summary panel so that mouse cursor changes to pointer when hovering over the arrows.

### DIFF
--- a/src/summary/ConditionSelection.css
+++ b/src/summary/ConditionSelection.css
@@ -31,6 +31,7 @@
 
 .arrow a:hover {
     color: #666;
+    cursor: pointer;
 }
 
 /* Disable arrow */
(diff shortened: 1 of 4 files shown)


Dtphelan1:
It does what we needed, but we really should be using classes and styling for this instead of inline styling. I'm okay to merge, but we should keep that in mind moving forward for all the other icon buttons that we'll need to use.

gregquinn2001:
It works and looks good. @Dtphelan1 raises a good point but let's merge as he says. Should we create a JIRA to go back and make the change to support this more generically going forward?

nicoleng12:
Good point @Dtphelan1 . I removed the inline styling, moved it to the css, and used a class. For the arrows in the condition selection, hover only works when the buttons are enabled. Because the buttons are disabled, the cursor will not change

Dtphelan1:
This looks fine for now!

PR #50, opened 2017-07-10
Shortcut viewer component

nicoleng12:
Created ShortcutViewer component which replaces the notes section in no patient mode. In SlimApp, added a currentDataItem field with a "shortcut" field. The shortcut field value is a string that is displayed in the ShortcutViewer component. currentDataItem gets passed from SlimApp to Shorcut Viewer. In Shortcut Viewer, it checks if currentDataItem is null. If it's not null, it displays the string value. If it is null, it displays a string that tells the user to pick a template from the left panel.

To test, run the app. By default the data model looks like this for currentDataItem:
```
 
currentDataItem: {
                shortcut: "place holder string for shortcut"
            }

```

For this case, the viewer displays "place holder string for shortcut". To test the null case I changed the data model to this:

`currentDataItem: null`

In this case the viewer displays "Choose a template from the left panel"

*NOTE: (See entry in null: https://www.w3schools.com/js/js_datatypes.asp). If currentDataItem is empty, as in it has no value:
 
```
currentDataItem: {
             
            }
```
then  when I do a null check, it still says that an object exists. It is an object that contains nothing so it fails the null check. I can completely remove the currentDataItem and then the null check works because the property doesn't exist so it's null, but based on the meeting this morning it sounded like we still wanted the property there just null. What I did instead was set the value to be null and then it evaluates to null in the null check. Let me know if there's a better way to handle this.

### DIFF
--- a/src/viewer/ShortcutViewer.jsx
+++ b/src/viewer/ShortcutViewer.jsx
@@ -9,19 +9,21 @@ import './ShortcutViewer.css';
 class ShortcutViewer extends Component {
 
     render() {
+        let string = "";
 
         if (this.props.currentDataItem == null) {
-            console.log("current data item is null");
+            string = "Choose a template from the left panel";
         }
         else {
-            console.log("current data is not null");
-            console.log(this.props.currentDataItem);
+            if(this.props.currentDataItem.shortcut) {
+                string = this.props.currentDataItem.shortcut;
+            }
         }
         return (
             <div id="shortcut-viewer" className="dashboard-panel">
                 <Paper className="panel-content trio">
                     <div>
-                        This is where the shortcut preview goes
+                        {string}
                     </div>
                 </Paper>
             </div>
--- a/src/viewer/ShortcutViewer.jsx
+++ b/src/viewer/ShortcutViewer.jsx
@@ -1,0 +1,32 @@
+// React imports
+import React, {Component} from 'react';
+
+// material-ui
+import Paper from 'material-ui/Paper';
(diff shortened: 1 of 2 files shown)


Dtphelan1:
I mean, I think the data model is a little disparate when it comes to what we discussed during the planning meeting. I haven't pulled it down yet, but is there anyway we could align what you're doing a little more closely with what we're going to have in terms of a long term model?

gregquinn2001:
I agree that would be nice. so change currentDataItem to currentShortcut and make it of type Shortcut (the abstract base class that Dylan created). You would then simply display the "choose from left" if currentShortcut is null otherwise you'd call getAsString() on the object and display that.

nicoleng12:
Very good points. I will merge what Dylan has so far into this branch and work off of his models

nicoleng12:
Merged Dylan's branch (shortcut-adt) into this one to use his data models for shortcuts. Leveraged Dylan's test button. On app start, viewer displays text for null currentShortcut. Clicking the button "test" at the bottom of the viewer triggers the function "changeShortcut" and sets the changed shortcut to progression. Because the state has changed and currentShortcut is no longer null, viewer calls getAsString on the currentShortcut object viewer and displays the string "Shortcut goes here"

*** Before merging, remove test button

gregquinn2001:
looks good. I'd say merge. It may be easier to let Dylan merge first though?

nicoleng12:
Yes Dylan should merge first just to be safe

PR #51, opened 2017-07-10
Shortcut adt

Dtphelan1:
As a subtask of task 291: No patient mode should replace editor with data item viewer and "Done" button

We've got a base class for shortcuts, an class called progression that extends shortcut, and in slimapp we have a shortcut tracked in state and a function for changing that shortcut. 

This feels a bit quick for my usual turnover, but it feels like this might be all we need in terms of modeling for task 291.  Can you all take a look at this, and if it looks good let's merge it to master; if you see things missing or think of other functions/features to add, let me know and I can get working on those.

### DIFF
--- a/lib/progression_shortcut.js
+++ b/lib/progression_shortcut.js
@@ -7,9 +7,33 @@ class Progression extends Shortcut {
         this.status = status;
         this.reason = reason; 
         this.observationDate = observationDate;
-        this.today = new Date();
         console.log(`constructor for a new Progression object`)
         // more Progression-specific stuff here, maybe
+    }
+
+    updateAttr(name="", value="") { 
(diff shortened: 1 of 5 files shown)


nicoleng12:
I am currently using your code in my branch and I think it's clear and meets the jira task. One thing: each of your case statements in SlimApp.jsx need a break statement

gregquinn2001:
I'd say merge this as well (after fixing switch per @nicoleng12's comments

nicoleng12:
I added breaks to the switch statements

gregquinn2001:
merge away!

nicoleng12:
Good to merge!

gregquinn2001:
I'm going to merge this so I can just work from it. I tested patient mode and it still seems to work as well

PR #52, opened 2017-07-11
New slimapp formtray

gregquinn2001:
This pull request addresses these items:
308 - Update FormTray to support new slim approach and legacy full app approach
287 - Implement progression button in "right panel" of no patient mode (calls changeShortcut)
304 - Keydates are shown in timeline as progression dates without labels

Details:
304 got fixed because I was figuring out how not to break timeline and saw the bug
287 got fixed because it was easier to do that with the already written changeShortcut method to test the FormTray switching between shortcut list and form for shortcut object
308 was the primary goal and the FormTray now accepts the Shortcut object which if null displays shortcuts which trigger a change in the current Shortcut. Once we have a non-null Shortcut, the component specified by the Shortcut getForm is displayed. Note that I removed updateAttr from Shortcut abstract base as it is not needed outside of the Shortcut object.


Dtphelan1:
Looks good to me to merge!

nicoleng12:
Same here, go ahead and merge!

PR #54, opened 2017-07-11
Tox Right Panel and Tox Shortcut

Dtphelan1:
For tasks 239 and 311: Implement Toxicity right panel and implement toxicity shortcut component, respectively. 

The changes are pretty straightforward: make a Toxicity class that extends the Shortcut class, and make a FormPanel for Toxicity. These are largely the only changes made, with the exception of a change from 'var' to 'let' in one file, and the move from !this.props.patient to Lang.isUndefined(this.props.patient) somewhere else, to improve quality of our code.


gregquinn2001:
Looks really good except the toxicity form needs to support adding and removing multiple toxicity values (grade and symptom). I like the component for searching symptoms!

we need a list with an add and remove button in the form as well. Only things in the list on the form will show up on in the shortcut phrase on the right as a comma-separated list in the brackets. So choosing grade and/or symptom alone will not update the shortcut in the shortcut viewer. Only once they then hit add. We can just support add and remove button for the list

nicoleng12:
This looks really good. The search component for symptoms is really nice. It's clean and easy to use. One thing, long symptom names like "Disseminated intravascular coagulation" gets cut off. Can we make the search text field wider? Whats' the longest symptom that could be displayed?

Dtphelan1:
Gotcha, working this morning on: 
1. Changing the data model to support a list of values; 
2. Including, as a first step, a + and - button for adding the current value and removing the most recently added value.

If anyone has a better concept of how remove should work, let me know

Dtphelan1:
Just made some changes to the data model and tested on my end. Let me know if this works and if not what else I can do  to improve. Again, this is simply meant to pass requirements for the current task; styling is not definite, but was sufficient to cover the required features.

gregquinn2001:
I don't think it works. As I covered earlier, I think there should be a list on the left side that you are adding to and removing from (based on selection within the list) and everything in the list is also shown in the shortcut. If I add 3 is it possible to remove the first one I added later with the current UI?

Dtphelan1:
Implemented a basic list view of current toxicities, support the selection of one such tox and the removal of the current tox if an identical one exists in the parents full list.

gregquinn2001:
basically functionality looks good. I'm going to create some jiras for some other work we can do here but let's merge this one!

Dtphelan1:
Sounds good!

PR #56, opened 2017-07-12
implemented copy button

gregquinn2001:
implemented copy button by nesting it within CopyToClipboard via react-copy-to-clipboard


Dtphelan1:
Works great on my end. Clear to merge

Dtphelan1:
Small list of features too, so I'm going to just push away and merge it. If I missedd something we can always undo it and talk more.

PR #57, opened 2017-07-12
Progression reasons without status

Dtphelan1:
For task 309: Progression shortcut only displays based on value if there's a progression value

Refactored the string creation to split up components into their own functions; and now when there are reasons but no progression status a '?' is displayed where the status would be.


gregquinn2001:
looks good to me. Merge it!

PR #58, opened 2017-07-13
Tox grade def improvements

Dtphelan1:
Based on task 314:  In Toxicity Form, swap order of grade and symptom and support grade definitions

To achieve this, we imported a more complete datamodel corresponding to the CTCAE (v4) based on an online xls file, filter the grades shown in the drop down based on the current adverse event, on choosing a new adverse event we remove the current grade selection if it's incompatible, and changed the representation of grades in the data model from a value (e.g. '1','2','3', ...) to the full text equivalent (i.e. 'Grade 1', 'Grade 2', 'Grade 3', ...).


Dtphelan1:
Added descriptions of grade based on either the current adverse event of based on the default description for grading

gregquinn2001:
awesome! merge it. Edwin is going to create some mock-ups based on our prototype to suggest how to improve the overall look and feel so I think it's great to demonstrate the concept

PR #59, opened 2017-07-13
Disable current button

Dtphelan1:
For task 315: In Toxicity Form, disable Add Current button unless both items entered

Pretty self explanatory. Does some checks to decide whether or not a current potentialTox can be added based on whether there is a potential tox and, if there is, if there are defined grades and adverse events.


gregquinn2001:
If I choose a symptom and then delete it, the button stays enabled and when I click it adds what I add selected previously and deleted (grade is selected throughout the above)

Dtphelan1:
Great catch -- I'll look into that!

Dtphelan1:
Check out these new changes!

PR #61, opened 2017-07-13
Fixed button width

Dtphelan1:
Based on task 316: Form panel buttons are different widths and should be the same size 

Improved the look of the form buttons by setting a (for now) sufficient width. Long term solution might incorportate a dynamic button sizing, but this should be sufficient for now.

### DIFF
--- a/src/forms/TemplateForm.css
+++ b/src/forms/TemplateForm.css
@@ -2,4 +2,5 @@
   float: left;
   clear: left;
   margin-bottom: 20px;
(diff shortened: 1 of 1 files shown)


gregquinn2001:
does what it says! merge it!

gregquinn2001:
can you pull this into sprint, give it story points, and assign it, and then move it to done! Thanks!

Dtphelan1:
Done and done

PR #63, opened 2017-07-14
disabled the addCurrent button if duplicate; don't allow duplicate adding in addTox fn

Dtphelan1:
Based on task 313: Toxicity Form allows adding duplicates

The add function blocks duplicates, and the addCurrent button won't be enabled if the value is a duplicate.


gregquinn2001:
looks like it works to me! merge it!

PR #64, opened 2017-07-18
Fixed increment of startDate bug

Dtphelan1:
Was working on integrating the progression shortcut object into the timeline and realized where the increment startDate bug was happening, as referenced in bug 278. Fixed it, and thought we might want to pull it in now.


gregquinn2001:
Yes we should definitely pull it into the sprint! we should merge.

PR #68, opened 2017-07-21
Removed case sensitivity in Adverse Event search in no-patient mode. …

mtorchio:
…Issue #318


gregquinn2001:
works. I was able to enter adverse event names in all lower case and all upper case and they both resulted in the grade list being updated correctly. Not sure if we want to leave the display to always have an initial capital letter??? Go ahead and merge though

PR #69, opened 2017-07-25
Current shortcut refactor

Dtphelan1:
Based on task 326: Modify patient mode to use currentShortcut.

Glaring differences: The inputs for the shortcut are not persistent, and dissapear the moment a new shortcut is generated. This is due to the lack of unified patient model. We could try to patch together some kind of read/write architecture based on what we talked about yesterday, but I thought we might want to wait until we started using the SHR models. This is up for discussion. 

Changes: 
- Removed several old functions/values used in updating TMN values, progression values, etc... 
- Removed any traces of two different shortcuts. 
- Change currentShortcut by typing into clinical notes, entering data into that shortcut as per usual.
- TMN values now no longer require a double click to input data. 
- Progression reasons are now trimmed for leading and tailing whitespace before validation. 
- Suppressed some logging.
- Data Summary Panel now displays based on currentShortcut, not the patient model passed. When we decide to publish data from currentShortcut to patient, this needs addressing. 

I think that sums up most of what's new. Let me know if you see anything else worth mentioning, any feedback on what's described above, and any suggestions regarding the patient data publishing. 

Thanks,
- Dylan

### DIFF
--- a/lib/staging.js
+++ b/lib/staging.js
@@ -1,10 +1,14 @@
 const lookup = require('./staging_lookup');
+const Lang = require('lodash/lang');
 
 // Computes the prognostic stage of breast cancer given T, N, M.
 // Uses the 7th staging edition, see:
 // https://cancerstaging.org/references-tools/quickreferences/Documents/BreastMedium.pdf
 exports.breastCancerPrognosticStage = (t, n, m) => {
   // Regardless of edition, metastatic cancer is always stage IV
+  if (Lang.isUndefined(t) || Lang.isUndefined(n) || Lang.isUndefined(m)) { 
+    return null;
+  }
   const M = m.toString().toUpperCase();
   if (M == 'M1') {
     return 'IV';
--- a/lib/staging_lookup.js
+++ b/lib/staging_lookup.js
@@ -172,13 +172,6 @@ exports.getTableForEdition = (ed) => {
 
 exports.isValidT = (possibleT, ed=7) => {
     const possibleTValues = getTsForEdition(ed);
-    console.log(possibleTValues)
-    console.log(possibleT)
-    for( const val of possibleTValues) {
-        console.log(possibleT.toLowerCase()) 
-        console.log(val.name.toLowerCase())
-        console.log(possibleT.toLowerCase() === val.name.toLowerCase())
-    }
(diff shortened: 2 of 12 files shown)


gregquinn2001:
looks pretty good. My only question is whether we can eliminate withinStructuredField within the FullApp and the associated methods handleStructuredFieldEntered and handleStructuredFieldExited.

Just changing currentShortcut should replace this behavior including setting it to null when the editor is not within a shortcut. This requires the editor to keep in its state the linkage between a # expression and the associated data object so that if they key into it, it can set currentShortcut to a shortcut that references the correct data object. We can merge what you have and create a new jira for this remaining work if you want

gregquinn2001:
plus FullApp and ClinicalNotes both have code specific to staging and progression. we need to eliminate that

gregquinn2001:
we should only have 1 place (probably in FullApp) that specifically references staging, progression, and toxicity as supported shortcuts

gregquinn2001:
maybe just merge this one and then I'll create a new JIRA

gregquinn2001:
Only SlimApp references progression and toxicity as a list of supported shortcuts passed into FormTray and then in the changeShortcut method. We need to achieve similar for FullApp

gregquinn2001:
jira 359 created in product backlog

Dtphelan1:
Great feedback -- yeah the minimal encapsulation still leaves something to be desired. I can take a look into some of this if we want. Though we might be able to tie some of those tasks in with the refactoring of Notes to use your new plugins.

Dtphelan1:
For now, I'll merge

gregquinn2001:
Yes totally agree may make sense as part of switching over to new editor and incorporation of new patient model. thanks!

gregquinn2001:
did you want to do more on 326 or can we move the overall task to done?

Dtphelan1:
Whichever you'd like -- If you want, for example, I can look into removing the withinStructuredField stuff as a first step. But as for the main task, I was done; just waiting to merge before moving to complete. Thanks for the reminder.

gregquinn2001:
you did merge??? let's leave the new jira task in the product backlog and focus on sprint items for now

Dtphelan1:
Yes, I did. I just didn't go to Jira right away. Hence the "thanks for the reminder".

PR #70, opened 2017-07-25
New shortcut plugin

gregquinn2001:
bring in TestApp which mirrors FullApp but uses new FluxNotesEditor instead of ClinicalNotes. Once TesApp has full functionality we'll switch FullApp to use FluxNotesEditor. goal of this sprint is to get there and end up with a more robust editor hopefully.


Dtphelan1:
Is there any reason that we added _getMostRecentProgression and _timeSorter to DataSummary? If not, I'd remove them for now and would be good to push on ahead.

gregquinn2001:
probably my bad. merge mistake. I take a look and commit changes. thanks

Dtphelan1:
Also, hold off on my approval for a second. I have some more feedback of bugs to log...

gregquinn2001:
bugs in what? If they are in the new editor they may be known issues. we're not saying this is done. we are trying to make it easier to work on it in parallel. that's why we are merging it as TestApp

gregquinn2001:
I removed those 2 functions from DataSummaryPanel

Dtphelan1:
I just thought we might want to log new bugs as we notice them. If all we're looking for is something to unify how we develop then yeah let's merge it. I just don't want issues to go-on noticed without being logged.

Dtphelan1:
Okay, barring the known issues that we'll log later, this is good to merge.

nicoleng12:
Ignoring look and feel, we can merge this. Im adding an issue to include the paper background for flux notes editor so that it is consistent with the other 2 panels

PR #71, opened 2017-07-28
new SHR-based patient model / Patient object / condition summary metadata

gregquinn2001:
initial commit of new patient model using SHR JSON instance with abstraction on top for accessing - Patient. Updated summary and timeline to use it and updated summary to use metadata defined in json per condition

### DIFF
--- a/src/notes/FluxNotesEditor.css
+++ b/src/notes/FluxNotesEditor.css
@@ -13,3 +13,112 @@
 	background-color: MediumBlue;
 }
 
+#clinical-notes #note-description p.note-description-detail-name {
+    padding-bottom:2px;
+    color: #919191;
+    font-size: .9rem;
+    margin: 0;
+}
+
+#clinical-notes #note-description p.note-description-detail-value {
+    padding-top:0;
+    font-size: .9rem;
+    margin: 0;
+}
+
+#notes-message {
+    text-align: center;
+    color: #00bcd4;
+    font-weight: bold;
+    font-size: 20px;
+}
+
+#data {
+    color: black;
+}
+
+.MyEditor-root {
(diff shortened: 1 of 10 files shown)


nicoleng12:
Adding back in the paper background to the editor panel makes the editor panel look consistent with the other 2 panels (and preserve the look and feel of full app). 

Insertion of data summary items into the editor works. I noticed, however, that there is an error when the user inserts the shortcut then inserts items from the summary panel. Inserted items from the summary panel turn into structured fields. 
image

Should we make a jira issue for this?

gregquinn2001:
I actually saw that too, but thought it would get solved once we can hook up the disable insertion when in a structured field. Dylan's changes making sure slate selection sync's with html focus will also help. I didn't create a jira. If after we implement that functionality fully, it still has issues, then we should enter them. I'll create a jira to implement the disabling of insertion buttons but I think it has to wait for your and Dylan's stuff to get merged with me stuff.

Dtphelan1:
I agree that it should probably wait until everything is merged together before we start knocking out those small, related tasks; they might be solved in the process of or as a product of fixing something in the existing tasks.

nicoleng12:
Ok based on your comments (Greg and Dylan) then this branch looks good to me

gregquinn2001:
Moved the hard coded metadata describing how to summarize a condition into a separate class (still hard coded for now though).
Also add focus to the transform chain after insertText for inserting summary items into editor. this resolves jira 213 so I pulled that into sprint

Dtphelan1:
Nice touch regarding the focus function.

gregquinn2001:
thanks! it's nice when things are that easy!

PR #72, opened 2017-08-02
Special keys behavior merged

gregquinn2001:
still have issues but wanted to allow merging of other features with this one. tab, shift-tab, F2, and Shift-F2 work for the most part. I will create new jira issues to cover what's remaining

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "react-flexbox-grid": "^1.1.3",
     "react-router-dom": "^4.1.1",
     "react-tap-event-plugin": "^2.0.1",
-    "slate": "^0.20.1",
+    "slate": "^0.20.1 <0.20.3",
     "slate-auto-replace": "^0.5.0"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


nicoleng12:
Tab and shift tab works. When the cursor is outside of the structured field and users enter tab, the selection moves from outside of the structured to into the structured field and traverses correctly from T to N to M dropdowns. 

F2 works, but only once the selection is in the structured field. For example, if the user types "abc", then inserts the shortcut and then places the cursor at the beginning of "abc"and presses F2, the selection does not move inside the structured field, whereas pressing tab will. Once inside the structured field (use selects a value in the T dropdown) and user presses F2, the selection correctly traverses from T to N to M dropdowns.

gregquinn2001:
yes, as I said, a lot of keys don't work 100% yet. I'm creating jira issues for them. we just want to be able to leverage the common infrastructure between this partial implementation and the work you and Mike are doing. We don't need to test different keys. They don't work yet.

PR #73, opened 2017-08-02
Publish value changes within structured field

nicoleng12:
Added handle update functions to publish value changes in the staging shortcut


nicoleng12:
Worked with Greg to merge master into this branch

PR #74, opened 2017-08-03
generic shortcut support (350) and current shortcut null if not in struct field (383)

gregquinn2001:
remove shortcut-specific code from everywhere except shortcuts, new shortcut manager, and a simple list of supported shortcuts in TestApp. also clear currentShortcut when not within a structured field in editor. also restore current shortcut to a structured field's corresponding shortcut when re-entering a structured field within editor. insert buttons in summary now correctly disable since currentShortcut is being set to null. Switches right panel to show form when in shortcut as it should. no update of drop downs in structured field from form changes yet. separate jira to cover that (368). added temporary button to insert each of the 3 supported shortcuts but the subfields aren't correct yet. other jiras to do those as well.

### DIFF
--- a/src/forms/FormTray.jsx
+++ b/src/forms/FormTray.jsx
@@ -52,21 +52,6 @@ class FormTray extends Component {
                 console.log(this.props);
                 panelContent = (
                         <DataCaptureForm
-                            // Staging Form data
-                            // Update functions
-                            onStagingTUpdate={this.props.onStagingTUpdate}
-                            onStagingNUpdate={this.props.onStagingNUpdate}
-                            onStagingMUpdate={this.props.onStagingMUpdate}
-                            onStageUpdate={this.props.onStageUpdate}
-                            // Helper functions
-                            calculateStage={this.props.calculateStage}
-                            // Properties
-                            tumorSize={this.props.tumorSize}
-                            nodeSize={this.props.nodeSize}
-                            metastasis={this.props.metastasis}
-                            stage={this.props.stage}
-                            
-                            // Progression data
                             // Update functions
                             changeCurrentShortcut={this.changeCurrentShortcut}
                             // Properties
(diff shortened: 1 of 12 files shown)


Dtphelan1:
Removed some antiquated code I found in the FormTray file, and separated the onF2 keypress to call an onF2 function (that just calls onTab). But all the changes you made looks pretty good. There are a few issues with the Dropdowns that we're already aware of. 

You're refactor looks good though!

Dtphelan1:
On second glance, we still hard-code the shortcuts in the temporary shortcut toolbar. We might want to pass in the shortcut manager and dynamically generate buttons based on information from the manager, but this might not be crucial since the toolbar is temporary.

gregquinn2001:
I guess the F2 thing is ok. The temporary toolbar is temporary. It will be deleted once Mike's change is in. It's fine the way it is.

Not sure I understand the 2nd paragraph. Can you explain it more?

Dtphelan1:
Okay, so no need to touch the toolbar. 

I thought that there were moments where the structuredFielPlugin was hardcoded (specifically line 524) but I realize now that what I was noticing was that ToxShortcut and ProgressionShortcut just haven't fully implemented their getStructuredFieldSpecification functions.

gregquinn2001:
I left the stuff in the Data Capture Form of FormTray as I wasn't planning to fix up the data capture stuff (curation functionality). Does it work after deleting those properties?

Dtphelan1:
It works as well as it did before (read: not well). These were very small changes that don't impact anything but took ~no-time-at-all to make as I was reviewing your code changes, so I thought I'd make them. Is that alright?

gregquinn2001:
anyone disagree with merging?

Dtphelan1:
Not at all

nicoleng12:
Running the test app, everything looks good. Nothing is broken (that we didn't already know about and already logged). Insert buttons are correctly disabled when in structured field and are re enabled when leaving the structured field. Right panel switches to the correct shortcut panel depending on which shortcut is active

The code looks good too.

Just noticed that when leave structured field and try to re enter via clicking on the dropdown, there is an error. Spoke with Greg and he just added a jira issue for it. 

Other than that good to merge

PR #75, opened 2017-08-03
Fix returning to structured field after leaving it (388)

gregquinn2001:
suppress onBlur event caused by giving focus to dropdown within structured field. added onOutsideStructuredFieldDropDownTab for handling tab key outside of dropdown. added data-offset-key attribute to select dropdowns to suppress many of the TypeErrors for cannot read property 'firstChild' of null resulting from content.js (slate) calling findDeepestNode based on a lookup of the above key. I left a lot of console.log messages in code for now. we will be debugging key behavior for a number of additional jira issues. It works to right arrow or tab from left side of structured field. It also works to just click on a drop down. not all keys that could return are implemented still and there are jira issues for them (e.g., shift-F2 or shift-tab when to right of structured field not implemented)

### DIFF
--- a/src/notes/FluxNotesEditor.jsx
+++ b/src/notes/FluxNotesEditor.jsx
@@ -68,6 +68,8 @@ class FluxNotesEditor extends React.Component {
     constructor(props) {
         super(props);
 
+		this.didFocusChange = false;
+		
         // Set the initial state when the app is first constructed.
         this.state = {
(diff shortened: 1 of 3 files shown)


nicoleng12:
Nice! This looks good on my end

gregquinn2001:
I merged it based on Nicole's comment. If there are other suggested changes, please add them as comments here so we can discuss and potentially create jira items

PR #76, opened 2017-08-04
staging updates between 3 places now (jira 357)

gregquinn2001:
changing staging in form updates dropdowns in current structured field. Also updates patient summary like it did in old app. added jira to address context (e.g., condition) and create a new staging per shortcut

### DIFF
--- a/lib/staging.js
+++ b/lib/staging.js
@@ -21,8 +21,8 @@ exports.breastCancerPrognosticStage = (t, n, m) => {
 
   // 7th edition staging, lookup[T][N]
   // null return values mean the staging is undefined for those T,N,M values
-  const ti = (lookup.getTsNamesForEdition(7)).indexOf(t.toString().toUpperCase());
-  const ni = (lookup.getNsNamesForEdition(7)).indexOf(n.toString().toUpperCase());
+  const ti = (lookup.getTsNamesForEdition(7)).indexOf(t.toString());
+  const ni = (lookup.getNsNamesForEdition(7)).indexOf(n.toString());
 
   if (ti === -1 || ni === -1) {
     // Urecognized T or N value
--- a/lib/staging_lookup.js
+++ b/lib/staging_lookup.js
@@ -35,9 +35,9 @@ const table5thEdition = [
 // 6th Edition, p.228:
 // https://cancerstaging.org/references-tools/deskreferences/Documents/AJCC6thEdCancerStagingManualPart2.pdf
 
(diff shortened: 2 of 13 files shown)


Dtphelan1:
On an irregular basis, I get a failure to sync that results in the following console error: 
At Patient.jsx:544, Uncaught TypeError: Cannot read property 'entryType' of undefined.

Dtphelan1:
It appears that certain combinations of TNM values don't work. Maybe this is intentional behavior?

gregquinn2001:
Any thoughts on how to reproduce? Could you post the entire stack trace?

gregquinn2001:
Yes, those combinations are invalid and show as undefined on the bottom of the right form. Should be very similar behavior to what we had before.

Dtphelan1:
Full description of bug: On the initial definition of Staging values, if an invalid selection of values is chosen, the link to the shortcut object stops functioning. Console continues to log an error in Patient.jsx:544.

gregquinn2001:
I fixed a few cases involving Tis and N1mi that weren't working correctly but no exceptions still

gregquinn2001:
let me see if I can reproduce that. thanks!

gregquinn2001:
awesome. reproduced it. thanks!

Dtphelan1:
Okay so I 've got the problem nearly pinned down. It has to do with the lack of a null check when you go to insert a staging value after lookup. The null value gets generated in StagingShortcut in updatePatient; the null value doesn't get checked in the call to patient.addObservationToCondition, which is definied in Patient.jsx:736

gregquinn2001:
I know how to fix it. I'm just on a call right now. I'll fix it after scrum probably

Dtphelan1:
Sounds good!

nicoleng12:
The summary panel, shortcut, and right form panel all sync nicely for staging. I see the bug that Dylan mentioned and it seems like you guys have a fix for it so that should be good. I'm sure you guys already know some of the invalid entries but in case Mike looks at this, an example invalid staging entry you can use to reproduce the bug is T2, N1mi, M0

gregquinn2001:
ok meetings finally ended at 12 so I was able to fix the problem and commit. let me know if you're ok with me merging before burndown

nicoleng12:
Tested and the bug seems to be fixed. Looks good on my end

Dtphelan1:
Bug has been fixed and warnings have been supressed, looks good to me!

PR #77, opened 2017-08-04
Autocomplete branch from master

mtorchio:
Enter still doesn't work but clicking does for suggestions.


nicoleng12:
Tested this and autocomplete works as well as clicking on suggestions. Just updated the package.json file to include slate-suggestions. Other than that, looks good to merge

PR #79, opened 2017-08-04
Remove shortcut specific code from autoreplace and suggestion

gregquinn2001:
use list of shortcuts passed in to initialize plugins. jira 401

### DIFF
--- a/src/notes/FluxNotesEditor.jsx
+++ b/src/notes/FluxNotesEditor.jsx
@@ -22,11 +22,11 @@ import StructuredFieldPlugin from './StructuredFieldPlugin';
 // Styling
 import './FluxNotesEditor.css';
 
-function getCurrentWord(text, index, initialIndex) {
+function getCurrentWord(text, index, initialIndex, initialChar) {
   if (index === initialIndex) {
     return { start: getCurrentWord(text, index - 1, initialIndex), end: getCurrentWord(text, index + 1, initialIndex) }
   }
-  if (text[index] === " " || text[index] === "@" || text[index] === undefined) {
+  if (text[index] === " " || text[index] === initialChar || text[index] === undefined) {
     return index
   }
   if (index < initialIndex) {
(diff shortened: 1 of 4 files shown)


gregquinn2001:
added generic support for inserters (name I used for the @ expressions). implemented name, age, gemder, dateofbirth, and patient for now

gregquinn2001:
still need to add autocomplete support them...

gregquinn2001:
autocomplete for @ inserters now supported

nicoleng12:
@ inserters work for all of the expressions (age, name, gender, dateofbirth, patient). 

When insert @ and then type out one of the expressions, prior to selecting the expression via enter key or mouse click, if the user deletes the expression back to the @ symbol, not all of the options are displayed again. For example, if the user types out @patient then deletes back to the @ symbol, the only option that is displayed is patient

gregquinn2001:
I think that behavior hasn't changed from when we switch to the slate suggestions plug-in. If you start typing you can get any of them. We should log this as a JIRA. It may be a bug in the suggestions plugin unfortunately.

gregquinn2001:
It may just be because the list is limited to only a certain number so it just picks that many to initially display when it only has the @ and no additional characters?

gregquinn2001:
actually I just played with it some more and it seems like if you type an @ after just completing one you only get the one you just used in the list. Once hitting a key, the complete list is shown (even alt-tab to switch away from the browser seemed to fix it). the suggestion plugin is not clearing the text used to match against. If I do it with age, I get age and stage only after typing @.

gregquinn2001:
added jira 426

gregquinn2001:
added jira 427 as well

nicoleng12:
Nice catch. 

Back in patient mode, when testing autocomplete, if you delete back to the @, no options are displayed. Options only show up once you type a character after the symbol. Not sure if we want to try to do the same thing and just remove display of options when it's just the @ symbol?

gregquinn2001:
probably makes the most sense, but either way it may require us updating the slate suggestions plugin. ugh

nicoleng12:
Since we captured the issues in jira I'm ok with this getting merged into master. I looked through the code and it looks good to me. I haven't done much work with the autocomplete though so I don't know if you want @Dtphelan1 to look at it before merging?

Dtphelan1:
So I'm pretty sure we ran into this issue when we were working on our plugin. We'll probably make a PR to the slate-suggestions. The issue is likely that they don't capture backspaces as a character that should be updating the possible suggestions. I anticipate that being the mechanism for resolving this bug. Yay.

gregquinn2001:
Dylan can you finish your review of this and approve or post additional issues to resolve?

Dtphelan1:
Sorry, got flagged down by Abhijay re: SHR.org questions. Reviewing now.

PR #80, opened 2017-08-07
Added functionality to switch title depending on if app is in patient…

nicoleng12:
Added functionality to toggle title. When in patient mode title is "Flux Notes". When in no patient mode title is "Flux Notes Lite"

### DIFF
--- a/src/nav/NavBar.jsx
+++ b/src/nav/NavBar.jsx
@@ -36,12 +36,21 @@ class NavBar extends Component {
   }
 
   render() {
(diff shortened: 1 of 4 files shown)


gregquinn2001:
I may have looked too quickly but let's make the title be passed into the NavBar by the specific App instance (FullApp, SlimApp, TestApp). NavBar should not have that information.

nicoleng12:
Sounds good. I can make those changes

PR #81, opened 2017-08-07
Revert "Added functionality to switch title depending on if app is in patient…"

nicoleng12:
Reverts standardhealth/flux#80


nicoleng12:
This needs to be rebased with greg's pull request that just got merged

PR #82, opened 2017-08-07
move TestApp to FullApp eliminating TestApp

gregquinn2001:
jira 400


nicoleng12:
@ expressions work and well as inserting shortcuts. Values are updated across all three panels (summary, notes, right form panel). Insertion of summary items work and timeline still works. This looks good on my end

PR #83, opened 2017-08-09
IE compatibility exploration

Dtphelan1:
This pull request has been put together after testing why our application wasn't functioning properly on Internet Explorer. 

Initial cause for issue was  ES6 JavaScript code in lookup files found in the /lib directory. Since these files were JS not JSX files, and since they weren't contained within the /src directory, they weren't being translated by Babel into ES5. This problem was solved first by changing the file extentions and moving the /lib directory underneath the /src directory, so that lookup files are now found in /src/lib/...

Second issue was due to polyfill and es5-shim issues; as you can see in last comment [here](https://github.com/facebook/react/issues/8379), load order matters when it comes to the polyfills and es5-shim with regard to the React and React-DOM. I've switched the load order and fixed this so that our application loads on IE.

Once that was remedied, there was an issue with the slate-suggestions plugin; specifically, the caret-position javascript file was relying on the use of window.scrollY/scrollX which is an invalid property on IE. Using window.offsetTop/offsetLeft respectively is a viable alternative. For the long term, I've lodged a [PR](https://github.com/oozou/slate-suggestions/pull/22) on the source GitHub repo. For the short term, I've modified the source files in node_modules, pulled them into our project and made the necessary changes. 

After making these three changes, the application works with minimal functionality on IE. The performance is severely slowly and ought to be optimized before running any initial pilots on an IE machine, but as far as testing functionality goes we have an alpha-grade build that works with IE. For this reason, I think the PR is ready to go through. 

Other Changes: 
- I've updated our version of react-scripts, the tool we use to build our project, from 0.9.x to 1.0.0. Consequently, developers should run an `npm install` after pulling latest from master. 
- Unified our usage of ES6 idioms throughout lookup files.
- Changed use of new.target to this.contructor in our shortcut file. This is related to the first issue and ES5 v. ES6. 
- Commented out log statements from feature-branches pushed into master to improve performance (this was done when to see if logs were the only weight holding back our IE performance; they were not.)
- EditorToolbar had an unnecessary anchor tag in one of it's render functions that caused a warning.  Removed that.  
- Added a few eslint-disable-line statements to the slate-suggestions module I pulled in to remove warnings; since this is not a permanent addition to the project I'm not worried about these changes.


gregquinn2001:
I'm hoping others can review this today, but a few quick thoughts:
(1) awesome job getting it working at all on stupid IE! :)
(2) did you already (if not can you) create a jira for IE 11 performance issues
(3) did you test the updated version on chrome and it still seems to work as well and with similar performance?

Dtphelan1:
1. Thanks!
2. Not yet, doing that now!
3. Yes and it works just as well as before. :)

gregquinn2001:
In the staging_lookup, why did you change Tis to TIS and change the comparison to include toUpperCase?

Dtphelan1:
Hm. Putting a hold on this PR... I'm having issues right now with the dropdown functionality.

gregquinn2001:
I changed it to return the correct format (Tis and N1mi) and do a case sensitive compare. That's what is in master right now. I'm  concerned about this pull request then. Not sure how that crept back in. I think we need to make sure it is just IE11 changes on top of master.

gregquinn2001:
It should be N1mi. Please diff your files with origin/master and make sure all the changes are ones you're making for IE11 maybe?

Dtphelan1:
I think I might have caused some unknown issues from the rebase. I'm going to check this a little more thoroughly, but I swear I did. Maybe my localhost didn't restart...

Dtphelan1:
This should run a lot smoother. I've walked through the diffs and all of them look intentional. I've also fixed a corner case issue regarding the clicking of the dropdown button; we now put focus back into the editor. Additionally, clicking the dropdown would occasionally cause log an error due to a return type in caret position. This might be due to the the plugin being 1-frame behind in editor updates. I've created a temporary fix that avoids the console error, but I think the suggestions plugin still has a few bugs that need addressing over time.

gregquinn2001:
the corner case you fixed only happened on IE11?

also, the error that's being logged from with the position code of slate suggestions has a jira of its own I believe already so no need to add one. Does your temp fix just suppress the error?

Dtphelan1:
It certainly doesn't solve it, I think; but it feels like a better solution than what's in place now. They were returning a null value, not checking the return value for null-ness, and then referencing a property directly. My fix returns an empty object, which seems to dodge the problem encountered by referencing a value on an object that doesn't exist. A better fix in my mind involves changing the plugin so it never runs into that scenario in the first place. 

The motivation for fixing this comes with the upgrade mentioned above from React-script v0.9.x, to 1.0.0. This new version results in a full-blown error page display when an error pops up in the console. The error page can be dismissed using 'esc' but it's alarming and less-than ideal; hence, the fix. 

If we want, we could always revert to react-script v0.9.x, but I thought it might be a good idea to use their first stable release in our project. Don't feel too strongly one way or another.

gregquinn2001:
yeah I diagnosed what was causing it as well. I was wondering if you just suppressed the error or actually put in a check for null in the adjustPosition function. I need that fix in my branch! That error is causing me problems!

I think an upgrade to any of our dependencies needs to be fully tested before making that leap. I think they typically deserve their own separate jira task unless you need it as part of your current IE11 task - then it certainly makes sense to do it now but changing a lot all at once is scary of course!

Dtphelan1:
Issues seemed to pop up on older versions of react scripts re: IE, and the switch was made in my (several) near-blind attempts to get something to generate on IE. Why don't I roll it back for now, see if it works with IE, and put a JIRA down for exploring the possibility of using a stable full-release of react-script?

Dtphelan1:
If you want, I can put a Null/empty check in there also. The fix I needed was simply to return an object (an empty one, but an object nonetheless) so I could try to reference a property it had.

gregquinn2001:
works for me if it works for you. Anything we can do to simplify what we're dealing with!

gregquinn2001:
For the null/empty check comment @Dtphelan1. I think we need to understand the situation in which position returns null (when the window selection rangeCount is 0) and what that means in order to figure out what it means to adjustPosition. The jira we have for that issue can do that. maybe add some comments to it? jira 384

Dtphelan1:
Sounds reasonable. For now I catch it an log more information, but I agree that we need to understand better what the plugin is doing and how it ends up in that position in the first place.

nicoleng12:
Awesome job getting this to work in IE! 

Tested this on IE and it works, but is a bit laggy in terms of performance. Also, user has to double click in drop down to get focus. Dylan already added jira items for both of these issues.

Tested this on chrome and all of the functionality still works.

gregquinn2001:
Thanks @nicoleng12! Any chance @mtorchio can test it out as well and do a quick check of code?

gregquinn2001:
@Dtphelan1  it looks like dev dependencies still says react-scripts 1.0.0 while the main dependencies section lists react-scripts 0.9.5. Is that what we want?

gregquinn2001:
It looks to me like you are re-adding ClinicalNotes.jsx in this pull request. That has been deleted in master. If you made changes in there, we'll need to figure out if those changes need to made in FluxNotesEditor or StructuredFieldPlugin

gregquinn2001:
wow. I went through all the changes. We need to see if we can standardize on white space so we don't have so many whitespace-related differences showing up. I did not compare FullApp since git diff just showed it as completely different. were there any changes in FullApp.jsx? Once we resolve these comments, I think we'll be good to approve. whew. A lot of changes are also due to commenting out console.logs. I wonder if there's a better way to handle debugging going forward? Can we do breakpoints? I know we can view values in the console if we could stop execution at a certain point with variables in scope.

Dtphelan1:
Three things: 
1. No we don't need ClinicalNotes.jsx. Again I think this was a quirk of doing my first rebase. I'll be sure to drop it first thing tomorrow.
2. I couldn't agree more on the whitespace. We should have a longer discussion re: styleguides/find one in the wild that works to solve these kinds of issues longterm.
3.  Console logging has gotten a bit excessive, I agree. One option is breakpoints, but that's a little tedious; another alternative is that we don't approve PR's if they still have PR specific console logs existing. This would have saved me a lot of changes re: logs statements from the selection debugging from a few days ago.

Dtphelan1:
If you have any more questions you'd like answered before approval, just let me know! :)

gregquinn2001:
need to address react-scripts dev dependency comment I made above as well

gregquinn2001:
What about FullApp as well

Dtphelan1:
We don't use dev dependencies right now, and react-scripts shouldn't be one anyway. That was an issue with our initial install. Removing now.

Dtphelan1:
No changes have been made to full app. Just checked both the old and new version side by side in my editor.

PR #84, opened 2017-08-10
Tab view for shortcuts

nicoleng12:
In Flux Notes Lite:
- Replaced forms panel with a list component (material ui list)
- List component dynamically creates list items based on string array passed into it from slim app
- Modified shortcut viewer so that in addition to the string shortcut result, it also displays the template for the shortcut selected from the list component
- Shortcut template functionality still works (selecting values updates string result correctly)
- Styled the layout a bit (following the Involution mockup) so that components sit in one Paper component as opposed to separate ones and played around with the width and height

Known things to work on (made sub tasks for this):
- Styling still needs to be done to more closely reflect the mockup (background color, etc)
- Styling still needs to be done on the shortcut forms to match the mockups


Dtphelan1:
This breaks (or erases) some of the changes made by my PR. We should review them and see how we can cherry pick your changes only into the project.

Dtphelan1:
Nevermind -- this branch just isn't rebased on the most recent version of master. We should do that before performing a full review.

Dtphelan1:
Performed the rebase. The rebase changes nothing content wise -- I checked before force pushing. Now time to check the branch fully.

nicoleng12:
Thanks for rebasing and pushing! I had rebased prior to making my pull request but did not do a force push. I will make sure to do that next time

mtorchio:
I ran this, and it looks great!

PR #85, opened 2017-08-10
jira 394 Context issue

gregquinn2001:
user can now use @condition to set context. StagingShortcut  and StagingForm now use SHR object TNMStage to track staging value. @stage can now be done but there must be a #staging in context.

If #staging is done without doing an @condition first, an error is produced.
If @stage is done without doing a an #staging first, an error is produced.

Still a point of discussion whether to handle it a different way

ContextManager is now used to keep track of context. Shortcuts now get the context manager when they are performing updates so they can update anything in context.

Staging values are correctly placed with patient object model now.

Progression and Toxicity need to be updated to do the same

ContextPortal could probably use some styling. I'll create a jira for that.


gregquinn2001:
added jira 435 for styling the ContextPortal

gregquinn2001:
added jira 436 for improved context handling: deletes and position dependent within document

Dtphelan1:
Putting focus back onto the editor after changing the conditionPortal state to active would be nice. Right now, after I choose condition as my new context, I have to click back onto the editor before the dropdown shows up. At least on my local version.

nicoleng12:
All the functionality that you added works on my side. 

The scrolling issue is back and is repeatable. When the user types @ or # and then uses the mouse to click on an option for the autocomplete, the page scrolls down. We had a jira issue for this, but I think we may have deleted this from our back log?

gregquinn2001:
Thanks Nicole. That issue is in our sprint backlog with 0 story points (397). I can't reproduce it though. I type # then use the mouse to choose an autocomplete and I don't see any scrolling.

Dtphelan1:
Ditto to Greg. Maybe try killing cloning the repo somewhere else on your machine and see if it doesn't work in that branch either.

mtorchio:
When I type @condition before #staging or #toxicity, I see the error in red text on the screen. 

This PR doesn't address #progression, does it? I'm able to type #progression without @condition and I don't see an error. However, trying to select items in the #progression drop downs gives this error in the console: 
Uncaught Error: update patient not implemented for ProgressionShortcut
    at ProgressionShortcut.updatePatient (Shortcut.jsx:59)

When I click an auto-complete suggestion, it scrolls to the bottom of the page for me as well.

nicoleng12:
I made a new clone of the repo and did not see the scrolling issue. I suspect that the issue may be due to different library versions. I am comparing yarn lock files to double check this

gregquinn2001:
Good catch Mike. Let me fix #progression at least. #toxicity is not implemented yet. I was originally just doing a POC for the context issue, but it should work with progression. 
Also, let me add an issue for exploring the "error handling"

gregquinn2001:
actually Julia is taking care of implementing for progression.  #progression doesn't work at all right now. It shows the staging drop-downs!

gregquinn2001:
Thanks Nicole. Let us know what you find out!

nicoleng12:
I did not notice anything different between the 2 files.... since the bug does not appear in a new clone of the repo, I vote for Mike and I making a new repo, working off of that, and merge this branch in. Thoughts?

gregquinn2001:
Call you call me via skype? I'm set up so just call me if you can

nicoleng12:
I spoke with Greg and we discovered that running yarn install instead of npm install fixes this. Mike can you try this as well?

gregquinn2001:
pushed commit with update yarn.lock that resulted from me running yarn install in this branch. updating wiki and project README to use yarn install instead of npm install

mtorchio:
I checked this out after the yarn.lock push into a new folder, and the scrolling issue has disappeared for me as well.

Dtphelan1:
@nicoleng12  or @mtorchio can you approve the changes if things look good on your end?

PR #86, opened 2017-08-11
Style lite-mode shortcut formList 

Dtphelan1:
The main purpose of this PR is to unify the styling of the lite-mode shortcuts dropdown with Edwin's mockups. This involved making some new css for slimApp, making FullApp's css more specific to impact just FullApp components and removing some unnecessary classNames. It also suppresses warnings from an earlier push.

Edit: Dropdown was a misnomer; the PR modifies the styling of lite-mode formList on the left hand side

### DIFF
--- a/src/forms/FormList.css
+++ b/src/forms/FormList.css
@@ -1,7 +1,8 @@
 #list-panel {
     background-color: #F3F3F3;
     /*Minus 64 for nav, minus 80 for search*/
-    height: calc(100vh - 64px - 80px);
+    /*height: calc(100vh - 64px - 80px);*/
+    height: calc(100vh - 64px);
     text-align: left;
 }
 
--- a/src/forms/FormList.css
+++ b/src/forms/FormList.css
@@ -1,7 +1,7 @@
 #list-panel {
     background-color: #F3F3F3;
     /*Minus 64 for nav, minus 80 for search*/
-    height: calc(100vh - 64px- 80px);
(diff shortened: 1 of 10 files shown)


gregquinn2001:
I don't understand I guess. what drop-downs exactly did you style? Do you mean the grade dropdown under toxicity and the progression status dropdown under progression? If so, why wouldn't the styling of those apply in full patient mode when they are shown on the right side?

gregquinn2001:
Also, I don't think we should have a search bar at all. We only have 2 shortcuts. Let's either just remove the dummy search bar or at least not include it if the number of shortcuts is less than 10 or something?

gregquinn2001:
We have a jira to implement the search already in our product backlog BTW

gregquinn2001:
should we clean up the console.log messages or are you thinking we may need them for some near future work?

nicoleng12:
Nice job on restyling Flux Notes Lite. LOVE the styling and how it more closely aligns with the mockups. Nice cleanup of warnings. I agree with Greg about the search bar. Can we hide that unless there are a lot of shortcuts in the menu?

Dtphelan1:
Dropdown was a misnomer. I restyled the list of shortcuts in no-patient mode. 

Sure we can hide the search for now. We wanted to align closely with the Mockups so I did the whole deal. If we want to bring the search back in when we implement the search functionality, we can do it then. 

I can drop the console logs.

PR #87, opened 2017-08-11
Create copy component

nicoleng12:
Created a new copy button that models the mockup. Copy button is initially disabled until user selects values from the shortcut template. The copy button text updates as the user builds the shortcut string (following the mockup design). Clicking on the copy button copies the shortcut string to the clip board. With Dylan's help, styled material-ui's raised button component to more closely match the mockups

### DIFF
--- a/src/forms/FormList.css
+++ b/src/forms/FormList.css
@@ -4,6 +4,8 @@
     /*height: calc(100vh - 64px - 80px);*/
     height: calc(100vh - 64px);
     text-align: left;
+    overflow-y: scroll;
+    overflow: overlay;
 }
 
 #list-panel .list-element.unselected{ 
--- a/src/forms/FormList.jsx
+++ b/src/forms/FormList.jsx
@@ -1,7 +1,6 @@
 // React imports
 import React, {Component} from 'react';
 // material-ui
-import Paper from 'material-ui/Paper';
 import {List, ListItem} from 'material-ui/List';
 // Styling
 import './FormList.css';
--- a/src/forms/FormList.jsx
+++ b/src/forms/FormList.jsx
@@ -1,6 +1,7 @@
 // React imports
 import React, {Component} from 'react';
 // material-ui
+import Paper from 'material-ui/Paper';
 import {List, ListItem} from 'material-ui/List';
 // Styling
 import './FormList.css';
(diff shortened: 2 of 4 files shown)


gregquinn2001:
it shouldn't be disabled initially. As the old version support, with nothing entered in the form on the right, the shortcut should simply be the #progression or #toxicity. The user should be able to copy that.

gregquinn2001:
other than that it looks great to me though. Did the mock-up dictate the amount of spacing between the form and the copy button? There seems to be a lot of space? Just asking.... not asking for a change.

nicoleng12:
Removed disabling of the button and reduced spacing between the copy button and the rest of the form

gregquinn2001:
Please don't add a subtask to redoing the form for that. It's not part of the POS forms. It something specific to no patient mode. Also though, it looks exactly like the mockup to me! What difference are you saying needs to be resolved?

Dtphelan1:
It's not the full width of the shortcutViewer, the centering isn't 100% right and the spaces between 'copy' and the element to be copied don't look 100% right. 

As for adding a subtask, you're right it shouldn't be a part of the POS forms. Separate components -- agreed. However, I'd like those styling concerns to be a part of subtask 431, "Style Flux Notes lite to reflect mockups. While we've made the left hand side look like the mockups, we haven't done that for the forms. That's what I meant to refer to.

Dtphelan1:
Made some small styling changes and cleaned up the shortcut viewer logic a bit. There were some dual-checks in place, some weird local vars, etc...

gregquinn2001:
The current jira is the one to address any discrepancies between mock-up and implementation. we shouldn't create a new jira; this one shouldn't close until we think it's right. I think it's good enough, but if we need it to be the same width then let's make it the same width before merging and closing this jira

gregquinn2001:
If the shortcut gets wider than panel, the copy button disappears. If it's easy to fix, let's fix it. If not, let's add a jira for it. Once toxicity is fixed to only support 1 at a time, we won't be able to produce this situation with current shortcuts

Dtphelan1:
Finished aligning the styling with the mockups. I'm merging since this had Greg's Approval and I'm adding a JIRA for the width problem since I didn't see it in the backlog. If we never run into it, all the better; but this way we have it.

PR #89, opened 2017-08-15
Left arrow functionality 

Dtphelan1:
Left and Right arrows can now be used to enter and exit structured fields, and to switch between dropdowns. It's known that the zero-width spaces at the beginning and the end of the structured fields require the user to hit left and right multiple time before entering and exiting, but this seemed acceptable based on conversations I had with @gregquinn2001.

### DIFF
--- a/src/notes/StructuredFieldPlugin.jsx
+++ b/src/notes/StructuredFieldPlugin.jsx
@@ -235,79 +235,40 @@ function onOutsideStructuredFieldDropDownEnter(event, data, state, editor, opts)
 }
 
 function onOutsideStructuredFieldDropDownLeftRight(event, data, state, editor, opts) {
-	console.log("onOutsideStructuredFieldDropDownLeftRight START.");
+	// console.log("onOutsideStructuredFieldDropDownLeftRight START.");
 	const { startKey } = state;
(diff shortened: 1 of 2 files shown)


nicoleng12:
Left and right arrows work to enter and exit structured fields with the already mentioned note that user is required to hit arrow multiple times.

When using the arrows to enter the structured field, there is the following error in the console:

image

Despite the console error, functionality seems to work. Once I use the arrow keys to enter a drop down I can successfully use the up and down arrows to select an options from the drop down, which also updates the right form.

Noticed that when in the structured form and click enter (for example when selecting one of the drop downs), a new drop down is created...Each press of the enter key creates a new drop down:

 
image

Dtphelan1:
1. The TypeError has been happening ever since we synched up selections. To be clear, it doesn't stifle functionality. 
2. Interesting bug. We can add a description of this oddity to JIRA 372. Again, to be clear, this doesn't mess up the functionality of this task right?

nicoleng12:
Yes, 1 and 2 don't break current functionality 

Just wanted to make sure the bugs were captured somewhere, but this branch does accomplish the task of using left and right arrows to enter/leave structured fields so I'm good with this being merged to master

PR #90, opened 2017-08-15
Enter click within dropdown now moves selection forward

Dtphelan1:
Clicking enter when focused on a dropdown moves you over to the next structured field. I haven't implemented a shift-enter functionality because it feels somewhat foreign to me, but if someone feels passionately that shift-enter should move you backwards, I'm happy to implement that feature!


nicoleng12:
Nice! Enter moves the selection to the next dropdown and gets rid of that strange bug where pressing enter creates new drop downs. 

One thing, can you remove the console logs you used for testing? Other than that, looks good!

PR #92, opened 2017-08-16
Removed support for multiple toxicities

Dtphelan1:
For JIRA 425: Remove support for multiple adverse events in 1 toxicity and update toxicity form. 

Front-end: 
- Users can select adverse events using a case-insensitive search.
- List of adverse events now include their associated descriptions, if they have a description. 
- List of adverse events are different bg-color and have a box-shadow. 
- Possible adverse events have a hover color. 
- Possible grade values are now represented with buttons, not a dropdown.
- Possible grade buttons have a hover color. 
- Possible grade buttons change bg-color and font color when selected. 
- Possible grade descriptions update when an adverse event is selected.
- Possible grades are filtered based on current adverse event.
- Users can click on the currently selected possible grade and un-select it.
- Users cannot click on invalid grades for a given adverse event. 

Back-end changes not shown through front-end features:
- Updated toxicity back-end to remove support multiple toxicities; toxicity is now a singular element., not a list of elements.
- Replaced Material-ui autocomplete with react-autosuggest, which allows for more complex option rendering.
- Removed pre-exisiting console logs and any that were created for debugging. 
- Cleaned up comment-descriptions of functions, tabs where present in files I touched, spacing, etc.. to align with new style guidelines.


Dtphelan1:
Note well -- we use a new yarn package, so make sure to `yarn install` before you test

gregquinn2001:
I still get compiler warnings about unused imports.

Dtphelan1:
Fixed that -- sorry!

nicoleng12:
Looks really good! Nice! When no adverse event is selected, there is still a listing of grades (in the mockup these are blank until user enters an adverse event). Which adverse event are the grades being displayed for when no adverse event is selected?

Can you make the grade row/box heights the same across all the grades?

gregquinn2001:
@nicoleng12  the grades are the generic definitions of each grade when no adverse event is selected. Do you think it's ok to show them initially? It actually allows a user to select a grade then an adverse event. It should not allow an invalid combination of course. I didn't realize the mockup forced selection of an adverse event first though. What do you we think is better?

Dtphelan1:
The mock up is wrong given our vision for these type of structured fields, similar to how your mockup had disabled buttons when it shouldn't have. I think users should be able to select a grade unless it's invalid, and it's not invalid until there is a conflicting adverse event.  

The same-height buttons isn't trivial, and I couldn't put together a quick fix. If you can/want, go for it; but I don't personally think it should block the PR. If we want to put a JIRA down I'd be good with that.

Dtphelan1:
I've got to get offline, but text me if you need me to change anything dire!

nicoleng12:
Looks good! I'll add a jira for the height

gregquinn2001:
I think saying the mock-up is wrong is a little strong. It's a different approach; it's certainly not incorrect.

not sure whose "mockup had disabled buttons when it shouldn't have"?

Anyways, @nicoleng12  merged it and created a jira for button heights.

Dtphelan1:
@gregquinn2001: I said the mockup was wrong "given our (current) vision for structured fields". That is to say, up until this point we had been enabling users to enter incomplete information and paste that into their notes, agnostic of what parts of the shortcut had been filled-in. If we're moving away from that and beginning to limit the kinds of information they can paste into their notes, that's a reason to blur the buttons until an AE is selected. But I wasn't aware of that shift. 

As for the mockup I was referencing, Nicole's PR for the [copy component](https://github.com/standardhealth/flux/pull/87) initially had a disabled copy button, as per the mockups; since we had been enabling the copying of partially-completed structured field phrases, we decided to enable it. But that was a previous case in which the mockups had a functionality for our slim-app structured phrases that differed from what we had envisioned.

PR #93, opened 2017-08-18
Landing page no patient mode

nicoleng12:
Created a landing page for Flux Lite: 

Front end:
- On app start, "Overview" page (the landing page) is shown is shown
- Users can always click to view the "Overview" page even if they click to a shortcut form
- landing page includes background information on the goal of Flux Notes Lite, information on SHR and a link to the SHR, as well as instructions for how to use the app
- Added helper text to the forms (modeling the mockups)

Back end: 
- Created a new form for the landing page that gets called when currentShortcut is set to null

### DIFF
--- a/src/forms/FormList.css
+++ b/src/forms/FormList.css
@@ -11,6 +11,9 @@
 #list-panel .list-element.unselected{ 
     border-bottom: 1px solid #d9d9d9 !important;
 }
+#list-panel .list-element.unselected.overview{
+    color: steelblue !important;
+}
 
 #list-panel .list-element.selected{ 
     background-color: white !important;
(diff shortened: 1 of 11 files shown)


gregquinn2001:
Looks nice. Here are a few items to think about:
[GQ1] I know we discussed calling it Overview but maybe the tab should be "About Flux Notes Lite" and then change the header at the top of the page
[GQ2] Remove the heading "Overview". I think the "About Flux Notes Lite" heading will be enough but then keep the "Using Flux Notes Lite" heading further down as is.
[GQ3] I think we should remove PATINA from the text and just say "capture of progression and toxicity. So I think we should add that we are allowing this as part of clinical notes.  I think the first sentence could be "Flux Notes Lite supports the ICARE data study capture of progression and toxicity during the authoring of clinical notes." ????

gregquinn2001:
I added Andre as a reviewer for this one as well.

nicoleng12:
I like your suggestions @gregquinn2001. I updated the page based on your comments (Q1, Q2, Q3)

gregquinn2001:
Looks good. I think Andre may try to get some language in there about enabling observational studies and allowing calculation of endpoints from treatment data similar to ones calculated as part of clinical trials. @andrequina ?

nicoleng12:
Merging this. Reviewed language on the landing page with Andre and he gave his approval

PR #94, opened 2017-08-18
Styling structured field

mtorchio:
This pull request includes a change to the structured field so that the styling isn't as "harsh". It uses react-select styled with CSS to create the dropdown menus in the structured field.


gregquinn2001:
drop downs are not maintaining values in all cases

mtorchio:
We will keep this branch for now in case we want to go back to the dropdown style of hashtags in the future.

PR #96, opened 2017-08-22
Locked material ui at version 1.0.0-beta.5

Dtphelan1:
As Greg said in his previous PR: "We use Material UI 1.0 beta 5 now. Yarn install will ask you to choose version from list. Always choose 1.0 beta 5 which is default right now. Can someone make yarn install just choose it? I didn't look into that..."

This updates the npm file and yarn file to look for material-ui v1.0.0-beta.5 explicitly.


nicoleng12:
Ran yarn install and no prompt for material ui version. Functionality of app still works

PR #98, opened 2017-08-23
Fix search field style in IE for patient mode

nicoleng12:
Removed height from ContextOptions.css file, which fixed the spacing issue in IE. Tested to make sure spacing of the search field doesn't overlap buttons in #progression, #toxicity, and #staging for chrome and for IE.

### DIFF
--- a/src/context/ContextOptions.css
+++ b/src/context/ContextOptions.css
@@ -1,9 +1,9 @@
 .btn_template_ctx {
-  width: 100%;
-  margin-left: 2px;
-  margin-right: 2px;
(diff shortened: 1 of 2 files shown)


nicoleng12:
Spoke with @Dtphelan1. The goal is to actually make the copy button text not centered (modelling the mockups). Made the fix so copy button text should be left justified in IE as it appears in chrome

Dtphelan1:
Confirmed that the left-alignment works on IE. My approval still stands -- thanks for adding that to the PR!

PR #100, opened 2017-08-25
Attribution on toxicity

jafeltra:
Added the attribution section onto toxicity. Added in Full and Lite modes. I talked with Mark on how to represent it in the SHR, and the valuesets are local right now but will eventually be updated I believe.

### DIFF
--- a/src/forms/ToxicityForm.css
+++ b/src/forms/ToxicityForm.css
@@ -69,6 +69,57 @@
     display: inline-block;
 }
 
+.button_selected {
+    background-color: #297DA2 !important;
+    color: #fff !important;
+}
+
+/*
+ * Tooltip Styling
+ */
+.tooltip {
+    position: relative;
+    display: inline-block;
+}
+
+.tooltip .tooltiptext {
+    visibility: hidden;
+    width: 150px;
+    background-color: #555;
+    color: #fff;
+    text-align: center;
+    border-radius: 6px;
+    padding: 10px 10px;
+    position: absolute;
+    z-index: 1;
+    bottom: 125%;
+    margin-left: 18px !important;
+    opacity: 0;
+    transition: opacity 1s;
+    font-size: .8rem;
+}
+
+.tooltip .large {
+    width: 220px;
+    margin-left: -95px;
+}
+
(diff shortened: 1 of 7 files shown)


gregquinn2001:
Looks great! In validateInCurrentContext within ToxicityAttributionCreator, if the first error occurs and you push the error message, you should return at that point. The 2nd error check requires a parent context which you don't have if the first error is true. If you copied this code from somewhere, maybe you could fix this issue there too? If what I mean doesn't make sense, let me know and I can just fix it.

gregquinn2001:
Should attribution be first on the form or last? I was expecting it to be after the adverse event and grade.

Dtphelan1:
@gregquinn2001 Good question regarding attribution. It looks like Edwin placed it at the top, but putting it towards the bottom might make more sense. Does anyone else have strong opinions on the matter?

laclark22:
@andrequina said he would like them at the bottom. I asked him after stand up when Greg made that comment.

jafeltra:
I will look into the error fix that Greg made, and I can make the two changes that Dylan mentioned. For the placement of attribution, I was just going by the mockups from Edwin, but I agree it makes more sense at the bottom, so I'll switch it.

gregquinn2001:
I think it makes sense at the end as it's not the primary value. @jafeltra can you move it to the end of the form and move it last in the list within ToxicityCreator.getValidChildShortcuts as well for patient mode.

jafeltra:
Yeah, I agree. I'll fix it in both those places.

Dtphelan1:
Re my require comment: It looks like the require statements started getting mixed into our code during the move a context driven approach. We should change them into imports on the files we touch in the future, but we shouldn't worry about changing all of them right now.

Dtphelan1:
Addresses what I was looking for, and moves the Attribution to the bottom of the page. Nice!

jafeltra:
Thanks! I also addressed Greg's comment and returned the errors array after the first error is triggered. I had just missed the return line, so the other files were correct and I didn't touch those. That set of commits addressed that, the import line, changing some tabs to spaces, and switching the order of attribution on the lite and full modes.

PR #101, opened 2017-08-28
Suggestion to insert now aligned with dropdown and text on page

Dtphelan1:
Item to be inserted on enter wasn't aligned well with dropdown and/or currently selected item in many cases. These changes should address that. 

To test: Try multiple permutations on the shortcut dropdowns and make sure the expected element is the one inserted. 

Additionally, the files in the slate-suggestion dir have been updated to use their pre-compiled ES6 src files, not the ES5 compiled files provided by NPM. This means that the files in flux\src\lib\slate-suggestions-dist should be identical with the files in [my fork](https://github.com/Dtphelan1/slate-suggestions/tree/fixed-one-off-isses/lib) of slate-suggestions. These will be the files I use to make a PR on [oozou/slate-suggestions](https://github.com/oozou/slate-suggestions).

Let me know if there is anything you see. If there are no stopping bugs,  I'd love to have this PR pushed through before 1:00 in time for a demo.


nicoleng12:
Tested this with drop downs for condition, staging, progression, and toxicity, Pressing enter correctly selects the item from the drop down. This looks good on my end

gregquinn2001:
When you choose from autoselect now, I don't get focus in the editor. When I insert via right panel or via typing, I maintain focus in editor. We can just log this as a jira and still merge this though if you want

gregquinn2001:
I'll add the jira. I assume Dylan is preparing for the demo

Dtphelan1:
I think we should merge it in and log the jira issue. For the record, I was also having issues shifting focus to the editor on master prior to the merge.

Dtphelan1:
Thanks Greg!

nicoleng12:
I already made the jira. feel free to add comments (issue 489)

gregquinn2001:
deleted mine as we did them at same time

PR #102, opened 2017-08-28
Position dependent context

gregquinn2001:
JIRA 459. Context is now dependent on location in editor. Any structured phrases that are contexts but appear after the current selection are not part of current active context.


mtorchio:
I think this is good, based on testing the context sensitive buttons on the right. So if the user types #progression   #Responding  based on  #Pathology  we still have the 4 remaining options visible (Imaging, Symptoms, Physical Exam, Markers). If the user only wants to select one of the options and move on to entering the Toxicity, they have to click the left-arrow at the top to make the condition name appear, then click it to see the buttons for Progression, Toxicity, and Staging. If they select all 5 options, it automatically goes back to the condition. This is okay because there's no way to know when they're done entering a Progression in cases where less than 5 options are selected.

In writing this comment I also noticed an error in copy-paste. I can copy both #Complete Response and #Responding without console errors but trying to copy #progression, #Pathology or the @condition causes this error:

base-64.js:21 Uncaught TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)

gregquinn2001:
thanks @mtorchio . There's already a jira to handle copy and paste that includes structured phrases in the product backlog. this jira was to make the right panels change based on location in the editor. If you move the cursor to before a context-creating structured phrase (e.g., @condition, #staging, #progression, #toxicity) then the tab for that context should no longer be available on the right and typing a sub-shortcut shouldn't work.

gregquinn2001:
It's actually not working at all now for some reason. I think something went wrong due to my git mistakes... please hold off for now on this pull request. sorry.

Dtphelan1:
Can do.

PR #103, opened 2017-08-28
Lookup now defines all grades as invalid for an empty AE

Dtphelan1:
Simple fix for removing grade value from toxicity when the adverse event is deleted. Original bug could be triggered in no patient mode by doing the following: 
1. Select adverse event; 
2. Select grade; 
3. Delete AE;

Bug: All grade buttons are disabled, but a grade is still selected on the back-end and can be copied into the users clipboard. This bug should no longer be the case.

### DIFF
--- a/src/lib/toxicity_lookup.jsx
+++ b/src/lib/toxicity_lookup.jsx
@@ -8800,7 +8800,7 @@ exports.isValidGradeForAdverseEvent = (possibleGrade, possibleAdverseEvent) => {
         return false;
     } else if (Lang.isUndefined(possibleAdverseEvent) || Lang.isNull(possibleAdverseEvent)) { 
         // Any grade is valid when there is no event
-        return true;
+        return false;
     }
 
     // If they are both


nicoleng12:
This looks good. Repeated steps listed in the comment above and deleting adverse event removes grade. Since this addresses the jira issue I'm good with this, but, do we want the attribution to be removed too when adverse event is deleted?

Dtphelan1:
That's a good question, one that I don't have a particularly strong opinion on. @gregquinn2001 what are your thoughts?

mtorchio:
I repeated the steps you mentioned and verified that the clipboard didn't include the deleted AE. This looks good to me. I think we probably should clear the attribution selection if the AE is deleted, because attribution is something that depends on the AE. If the user has just copied the string and wants to use the page to generate another string, deleting the AE should reset the page to its original state. It's clearer than a Reset Page button or clicking the browser refresh.

gregquinn2001:
I don't think attribution should be removed as well. grade and adverse event are related so it makes sense I guess although our previous version allowed selection of grade first - not sure why we went away from that!? I actually preferred the old method that allowed selection of grade first.  If the form starts out with disabled grades until you select an AE, then going back to no AE should result in disabled grades. attribution is always enabled. All other forms don't disable anything that I'm aware of. you can select the attributes in any order.

Dtphelan1:
I guess I'm a little confused because that was my preferred behavior as well, as discussed in detail on [PR 92](https://github.com/standardhealth/flux/pull/92). 

The reason we moved away from it was that changes made in [PR 95](https://github.com/standardhealth/flux/pull/95), specifically in [this commit](https://github.com/standardhealth/flux/pull/95/commits/21a8666729a50613213916625a16d2b0f5252c15) disabled grades when there was no adverse event. There was no response to the last comment on PR 92, a comment made after 95 was merged, nor was there a response to the comments I made on PR 95 directly regarding DP1 -- with that in mind I assumed we were just moving forward in the direction of disabled grades.

gregquinn2001:
If I remember correctly, someone questioned the disabled and the author's response was that the mock-ups showed disabled buttons for the grades. We should follow the mock-ups and see if any clinician feedback leads us to change it? Any other thoughts?

Dtphelan1:
Sure! I'm fine leaving it as-is for now and waiting for some clinician feedback.

PR #106, opened 2017-08-29
JIRA 459 Position dependent context

gregquinn2001:
Now the contexts showing in the right panel are dependent on the location within the editor. If the cursor is before a # expression that is a context, that context won't be active.


laclark22:
Looks good to me. Just thinking out loud... when you press enter and get to a new line does it make sense to default back to the #staging, #toxicity, and #progression option? Right now it just stays on whatever options are associated with the last button you added.

gregquinn2001:
interesting idea. We could explore that. So then context would be active for the paragraph where it started only?

I feel like there could be some strange corner cases, but trying to think through it, I can't see them. It's not immediately obvious to me how to implement it though. I guess only descendant blocks from the current paragraph block would have potential to be in context instead of the entire document... Maybe that could work

jafeltra:
This mostly looks good to me. I was able to see the context changing as you moved the cursor back to different contexts. There was one bug I noticed, although I've only been able to get it to come up in certain cases. So it might be fine to merge this without fixing the bug and creating a Jira item for the bug.

JA1: When entering information with a combination of typing and using the buttons on the context panel, information in the panel disappears as if it is being entered, but it is not. One way to reproduce this:
1. Type @condition and choose Invasive...
2. Click the buttons in the context panel for "Staging" and "T0"
3. Use the tabs to go back to the "Invasive.. " context and use the buttons to choose "Toxicity" and "Anemia".
4. Move the cursor back to after "T0". This updates the context in the right panel, as I expect it would and as this pull request addresses.
5. Use the buttons on context panel to click on "N0". This makes the N values disappear from the context panel as if they were entered, but nothing is updated in the editor. If you select the M value, nothing happens but then the context form resets and you can successfully click to enter data.

I have noticed something similar in some other cases, but I haven't been able to reliably recreate them. Again, this seems small so I'm fine if it gets separated from the pull request. I only noticed it as I was trying to partially enter a context and then go back to it to finish entering.

Dtphelan1:
I haven't seen these bugs on my machine. Maybe we can make a jira for it now and merge this in?

jafeltra:
Okay that works for me! Maybe it's just doing something weird on my computer. Especially if you don't see them, I'm good with merging it.

jafeltra:
I just talked to Dylan about this and we verified the bug. It is caused by not adding in spaces after the shortcut values. I'm going to create a separate JIRA issue for this, but the pull request looks good to me!

PR #109, opened 2017-08-30
JIRA 491. onclick handler on paper to give focus to editor

gregquinn2001:
When focus is not in editor, clicking any line in editor down to bottom of paper should give focus to editor


nicoleng12:
Looks good on my end. Straightforward code fix.

PR #110, opened 2017-08-30
JIRA 494: Context portal now sensitive to pageXOffset and pageYOffset

Dtphelan1:


nicoleng12:
Looks good on my end! Nice

PR #114, opened 2017-08-31
Unify yarn and package

Dtphelan1:
yarn.lock and package.json are now synced to install the same files.


gregquinn2001:
do we need to clean up yarn.lock though? It seems to have a lot of dependencies in there and I'm thinking some of them are not used??

Dtphelan1:
This is a good point, though I think most of those dependencies you're seeing in yarn.lock are just the child dependencies of packages we do use. There were a few packages in our package.json file that we don't use anymore, though. So good catch!

gregquinn2001:
ok great. good to clean it out while you're in there. thanks!

gregquinn2001:
Note that you didn't add anyone as reviewers so that may be why no one else has looked at this. I added myself as a reviewer

Dtphelan1:
whoops....

PR #117, opened 2017-09-05
Context menu styling

jafeltra:
This adds styling for the context panel and adds tooltips to the buttons on that panel. In order to do that, we had to reformat the getTriggers() methods to return an array of objects instead of strings. I think that I've caught all the cases where this broke things, but if anyone notices any that I missed when reviewing this, let me know!

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "moment": "2.18.1",
     "positions": "1.6.1",
     "prop-types": "15.5.10",
+    "rc-tooltip": "^3.4.8",
     "react": "15.5.4",
     "react-autosuggest": "9.3.2",
     "react-calendar-timeline": "0.14.0",
--- a/src/context/ContextOptions.css
+++ b/src/context/ContextOptions.css
@@ -9,6 +9,7 @@
 .context-options-list {
     overflow-y: scroll;
     overflow-x: hidden;
+    overflow: visible;
     height: 710px;
 }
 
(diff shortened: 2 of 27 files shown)


gregquinn2001:
Looks good overall. The only issue I see is that when I click on a button in right context panel when a tool tip is displayed the updated context panel will have a tool tip displayed right away. Maybe on click needs to close the tool tip if it's displayed?

gregquinn2001:
Also, why do we have getName on StagingCreator and ToxicityCreator and getShortcutGroupName on their other shortcuts? All shortcuts have getShortcutGroupDescription though? Why don't we always use getText and just have a getDescription? Isn't getDescription describing the shortcut itself?

nicoleng12:
[NN1] - similar to Greg's comment. There is a tooltip displayed when the context panel updates. I see this even when no tooltip is displayed before updating the context panel. For example, I click @condition and choose "Invasive ductal carcinoma of breast". Then I select #staging. When the context panel updates, there is a tooltip already being displayed for M0. 

[NN2] - The buttons for #progression in the context panel overflow the panel

image

[NN3] - The padding for the toxicity context panel is much smaller than staging and progression

image

versus

image

jafeltra:
Thanks for the reviews everyone! I had noticed that tooltip continuing to display and forgot to look into why it was happening, so I'll do that now! I think that NN1 is related to that issue. NN2 is related to previous issues with the tooltip that I forgot to undo, so thanks for noticing! NN3 I think might have been styling that Laura was looking at. I probably should have double checked with her before making the pull request. I had thought she had finished, but those styles should be fixed. I might wait on NN3 for tomorrow to see if Laura had looked at it.

jafeltra:
Also, regarding Greg's second comment, I believe that getShortcutGroupDescription can actually be removed. I hadn't realized it until you pointed it out and I thought about it again. That function was being used to associate the parent group's description with a child shortcut (e.g "Describes the degree to which the cancer has reached nearby lymph nodes" for "N" values). However, that information is not being written on the panel anymore since we are just writing "Node Size" above the group of N values now based on feedback that the description took up too much space. That description is not used anywhere now (in a tooltip or directly on a panel). If that is something that should be put somewhere, let me know. I'm just not sure where it would go.

getShortcutName provides that "Node Size" description since that is not saved anywhere in the data. This is different from what getText returns because we want a description instead of the actual trigger that would be typed. As for why some are getShortcutName and others are getName, I'm not sure. Those are something Laura wrote and she might have a better answer. I would guess the names can be changed to all align.

gregquinn2001:
ok. sounds good. I like removing getShortcutGroupDescription but I think they should all have a getDescription. I'm guessing the issue is with the T shortcut for example. You have getDescription which returns the description for the current value, but you also want a description for t values as whole. I think getDescription should describe T values as a whole since that's what the class is for. The description for the current value should be getTextDescription or getValueDescription - whichever makes sense.

The name method makes sense but hopefully Laura can resolve the 2 different names.

nicoleng12:
[NN4] - After talking with Julia, noticed that button widths are not consistent across panels. For progression and staging, button widths are constant but for toxicity, the button widths are dynamic (fitting the text). This should be consistent across all the panels. Should the button widths be the same regardless of the length of text or dynamic so that the button width just fits the text?

gregquinn2001:
For NN4, getting it to work the way it does right now was a lot of work. We can try another round of improvements but getting it better may be a huge challenge. Not saying we shouldn't try, but maybe we do it as part of the next round of changes when we introduce the vertical push buttons instead of the tabs?

nicoleng12:
For NN4, yes I'm ok with adding it as part of issue 497

jafeltra:
I just updated this pull request. The updates fix the tooltip errors that I believe people were seeing. They also fix the naming inconsistencies that Greg pointed out. NN2 is also fixed. I believe it was decided that NN3 and NN4 were related to each other and will be looked at in future tasks, so they are not fixed here.

PR #118, opened 2017-09-07
Restyled Flux Notes Lite About page to match suggestions from Edwin -…

mtorchio:
… space before Using Flux Notes Lite and numbered list left-aligned

### DIFF
--- a/src/forms/LandingPageForm.css
+++ b/src/forms/LandingPageForm.css
@@ -1,3 +1,10 @@
 .landing-page-title {
     color: steelblue;
 }
+ol, ol li { 
+    margin-left: 0; 
+    padding-left: 0; 
+}
+ol { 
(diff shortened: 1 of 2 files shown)


nicoleng12:
Looks good!

gregquinn2001:
It looks like the numbered list is not left aligned with the other text on the about page. Is this intentional? Shouldn't they be left aligned?

mtorchio:
Greg, this is something that varies slightly between browsers. The numbered list is very slightly indented in Chrome, but lines up neatly in IE. To align it neatly in Chrome, we could decrease the ol margin-left property to 1.0em, but then it would be outdented in IE. Since this page is more likely to be viewed in IE and slight indenting looks better than slight outdenting, I set it to 1.3em.

PR #119, opened 2017-09-07
Spacing between selections and fixed copy button text

nicoleng12:
- Added spacing between selections on the toxicity and progression panels for lite mode (spacing on the landing page is being handled as a part of jira issue 

- Edited helper text for copy button

### DIFF
--- a/src/forms/ProgressionForm.jsx
+++ b/src/forms/ProgressionForm.jsx
@@ -99,7 +99,7 @@ class ProgressionForm extends Component {
                 </p>
                 <Divider className="divider"/>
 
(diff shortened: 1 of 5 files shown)


gregquinn2001:
Spacing looks good. I'm still struggling with the prompt though.

What about something like this for the prompt:
"When finished selecting values, click on the copy button above and then paste into a note within your EHR."

jafeltra:
I think I agree with Greg. I like the mention of clicking a copy button. But the spacing looks good to me too!

nicoleng12:
I updated the text

PR #120, opened 2017-09-08
Insert summary items via structured phrases

gregquinn2001:
Solves 3 JIRA items:
510 - Modify insertion from summary to use shortcuts. insertion now uses NoteParser as well
453 - fixed bug where clicking next to a missing item in summary where + insertion button would normally be would cause insertion of "Missing x" even though plus button wasn't visible.
496 - if context is fracture condition (any non-breast cancer condition actually) then #toxicity, #progression, and #staging should not be valid

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "moment": "2.18.1",
     "positions": "1.6.1",
     "prop-types": "15.5.10",
+    "rc-tooltip": "^3.4.8",
     "react": "15.5.4",
     "react-autosuggest": "9.3.2",
     "react-calendar-timeline": "0.14.0",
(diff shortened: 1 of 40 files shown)


jafeltra:
This looks good. I just have one question. Should the values under Progression on the summary panel also insert shortcuts? I'm not sure if they're different for some reason that I don't know about so I just wanted to double check.

gregquinn2001:
good question. I should have addressed that in my pull request description probably. I added the ability for summary items to insert shortcuts, but I only changed the first two to actually do it because those are the only ones I saw in a quick look that had an inserter shortcut corresponding to what they are inserting. we don't have any support right now for @progression or @progression status or @progression reasons....

jafeltra:
I definitely didn't register that difference. Thanks for pointing that out. That makes sense then! Looks good to me!

nicoleng12:
Inserting from summary correctly inserts the shortcut, clicking next to missing items no longer inserts into the notes field, and if context is fracture, toxicity, progression and staging are not options if the context is fracture. Looks good!

PR #122, opened 2017-09-08
No spaces in search

jafeltra:
This makes it so that any spaces (extra or none) that a user enters in the search bar on the toxicity form in Slim Mode are ignored and returns values either way. (Ex: "bonemarrow", "bone marrow" and "    bone    ma rrow" all give a result of "bone marrow"). This does not change the search bar on the context panel in Full Mode. I will make a separate Jira item for that.


nicoleng12:
This looks good! Nice job

PR #123, opened 2017-09-12
Made copy button a fixed at the bottom of the page so that it doesn't scroll

nicoleng12:
Tested this on both chrome and IE. Please make sure to test on both browsers when checking this pull request.

Note: css `position: sticky` would have easily solved this jira issue but it is not supported by IE so ended up having to do some styling with css and `position: fixed`. 

Changes made:
- Fixed the position of the copy button component
- Added background color and shadow border to div containing copy button component
- Added extra margin bottom to the panel content component so that it doesn't get overlapped by the copy button component

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -2,9 +2,9 @@ body {
   margin: 0;
   padding: 0;
   height:100%;
-
   font-family: 'Roboto', sans-serif;
 }
+
 ::-webkit-scrollbar {
     width: 12px;
     height: 12px;
--- a/src/viewer/ShortcutViewer.css
+++ b/src/viewer/ShortcutViewer.css
@@ -2,8 +2,8 @@
 #shortcut-viewer {
     background: white;
 
-    /* 64 fo navbar; 60px for top/bottom padding */
-    height: calc(100vh - 64px - 60px);
+    /* 64 fo navbar; 60px for top/bottom padding; 104px for copy button */
+    height: calc(100vh - 64px - 60px - 104px);
     text-align: left;
     z-index: 2;
     padding: 30px 60px;
--- a/src/viewer/ShortcutViewer.css
+++ b/src/viewer/ShortcutViewer.css
@@ -19,21 +19,22 @@
 
 #copy-component {
     position: fixed;
-    bottom: 10px;
+    bottom: 0px;
+    left: 25vw;
+    width: calc(75vw - 12px);
     background-color: #ffffff;
     border-top: 1px solid #d9d9d9 !important;
     text-align: left;
-    -webkit-box-shadow: 0px -4px 3px rgba(125,125,125,0.7);
-    -moz-box-shadow:    0px -4px 3px rgba(125,125,125,0.7);
-    box-shadow:         0px -4px 3px rgba(125,125,125,0.7);
+    -webkit-box-shadow: 0px -2px 3px rgba(125,125,125,0.7);
+    -moz-box-shadow:    0px -2px 3px rgba(125,125,125,0.7);
+    box-shadow:         0px -2px 3px rgba(125,125,125,0.7);
 }
 
 .btn_copy {
     display: inline-block;
     min-width: 96% !important;
-    margin-top:  20px;
-    margin-bottom: 20px;
-    margin-left: 20px;
+    margin: 20px 32px 20px 20px;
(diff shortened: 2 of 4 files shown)


gregquinn2001:
looks good and works quite well in Chrome (I haven't tested in IE yet). Can we make the scroll bar end at the top of the fixed portion though? On some platforms, the scroll bar could end up behind the copy button area and not be accessible via the mouse - plus we aren't scrolling that part so should it be part of the area next to the scroll bar?

gregquinn2001:
looks the same on IE. nice job!

jafeltra:
This looks good to me too! I did notice the same thing as Greg where it seems that the form continues behind the copy button, so the scroll bar gets cut off and some of the text can be seen under the copy button.

jafeltra:
After looking at this a little, I had an idea about how to fix the scroll bar/text behind the copy button. I can make the fix if Nicole seems to be offline the whole day, but I figured I'd post it here to get feedback and see if other people like the idea. 

In ShortcutViewer.css, in #shortcut-viewer, the height is being calculated. I thought we could subtract out the height of the copy button as well, since it is already taking into account the nav bar and the padding. This would make the height line look like:  

/* 64 fo navbar; 60px for top/bottom padding; 104 for copy button */
    height: calc(100vh - 64px - 60px - 104px);

gregquinn2001:
sounds good to me. I'd say go for it and commit to @nicoleng12's branch. I don't think she's actively working on it right now.

jafeltra:
I pushed the change. It fixes the comments that were made on this pull request.

gregquinn2001:
do you want to approve it then Julia?

gregquinn2001:
actually it looks a few pixels off maybe? The look of the top of the scroll bar when it is up against the top does not match the look of the bottom of the scroll bar when it's at the bottom. 
Also, is there a lot of extra space at the bottom of the form intentionally right now to account for the space under the copy button? Can we reduce that? Wow, I feel really nit-picky! sorry!

jafeltra:
Ohh I didn't even notice that it changed the space on the about page. This might not be the right fix in this case. But I can look into it.

gregquinn2001:
good catch!

jafeltra:
Yeah! Definitely! I can revert the commit if we want.

gregquinn2001:
Also, it seems to have caused the copy button to extend off the screen to the right now? I checked in master; it doesn't do that.

gregquinn2001:
If you are going to stop working on it then I'd say undo your commit; otherwise, up to you.

jafeltra:
I undid the change that I had made earlier. I don't think that it was the correct way of making the fix and caused additional problems. I'm going to focus on looking at a different task now instead of looking for a better solution to this. Nicole may have other ideas, and this is demo-able as it is now.

nicoleng12:
Played around with this and didn't quite find the perfect solution yet. I didn't commit it because it doesn't look quite right, but if you want to try what I did I'll list out the steps here. In trying to address the copy button covering the scroll bar, instead of adjusting the height of panel content (which Julia and tried and I also tried but couldn't get it to quite work) I was thinking of just making the width of the copy button narrower so it doesn't cover the scroll bar.

What i did:

- In ShorcutViewer.jsx, in the return inside the render method, move `{copyComponent}` to inside the div with the id "shortcut-viewer"
- In ShortCutViewer.css, remove the width in copy-component

The result will look something like this:

image

The next part would be to fix the width of the copy button so that it lines up properly

nicoleng12:
I pushed some changes and also rebased off master. I was able to narrow the width of the copy button and made sure the button and the button div line up correctly. Let me know what you guys think of this approach for fixing the copy button so it doesn't overlap with the vertical scrollbar. @gregquinn2001  @jafeltra

Dtphelan1:
I took a look at your changes Nicole and made a few minor styling tweaks. It's not perfect (especially on small screens) but I wanted to help how I could in order to facilitate the push to master. Let me know if these changes are helpful. If not, we can revert no problem!

nicoleng12:
Thanks, this looks great Dylan! It matches the mockup and doesn't overlap with the vertical scroll.

jafeltra:
This looks great! Nice work Nicole and Dylan!

PR #124, opened 2017-09-13
Progression date assessed

jafeltra:
The goal of this pull request is to add the "#as of #13 Sep 2017" string automatically to the copy line in slim mode. There is no way to adjust the as of date. The "#as of" shortcut is not created yet either. Although it does have some functionality in full mode, the goal of this task is not to have that fully functioning yet, so not everything that will be needed there is implemented.

### DIFF
--- a/src/patient/Patient.jsx
+++ b/src/patient/Patient.jsx
@@ -853,6 +853,10 @@ class Patient {
         });
         progression.evidence = reasonObjects;
     }
+    
+    static updateDateForProgression(progression, date) {
+        progression.clinicallyRelevantTime = date;
+    }
 
     static _progressionStatusToCodeableConcept(status) {
         if (status.toLowerCase() === 'complete response') return { value: "C0677874", codeSystem: "http://ncimeta.nci.nih.gov", displayText: "Complete Response"};
--- a/src/patient/Patient.jsx
+++ b/src/patient/Patient.jsx
@@ -854,7 +854,7 @@ class Patient {
         progression.evidence = reasonObjects;
     }
     
-    static updateDateForProgression(progression, date) {
+    static updateAsOfDateForProgression(progression, date) {
         progression.clinicallyRelevantTime = date;
     }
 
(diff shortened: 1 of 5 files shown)


mtorchio:
Ran and tested it, the date is present and can be copied with the rest of the progression text. Looks good to me.

gregquinn2001:
This looks great. A few nits (only the first is really part of slim mode anyways):

Please rename updateDateForProgression to be updateAsOfDateForProgression.
DateCreator.getShortcutType should return "#date"
DateCreator.validateInCurrentContext should just return an empty list since DateCreator is generic
DateCreator.getDescription should just return "a date" I guess. It's not always today's date.

gregquinn2001:
In ProgressionCreator the attribute is just "date" right now. Let's change that to "asof" or "asofdate"

jafeltra:
Ah these are all good points that I didn't think to incoporate after deciding that Date was going to be generic. I can fix them! For the DateCreator.validateInCurrentContext comment, should the method be returning an empty list for errors? I'm just clarifying what you meant. Also, I'm thinking that eventually this should be changed to check if the context is "asOf" or "referenceDate", but since those don't exist yet, I won't add them. I can add a note though if that sounds right.

gregquinn2001:
Yeah sorry. good clarification. DateCreator.validateInCurrentContext should always just return an empty list of errors because it is valid in any context. The user can always specify a date. the parent context will receive that date and it will in turn pass it on to its parent context knowing which attribute to put it in. I think that should work. So, the DateCreator will not know anything about who can be its parent as it is a generic way to enter a date in a note as structured data. So I don't think it should ever check if its parent is asOf or referenceDate. Does that make more sense?

jafeltra:
Thanks for the clarification. I think that makes sense. I'll leave it as an empty list of errors.

jafeltra:
I just updated the pull request based on Greg's comments. Let me know if any other updates are neeed!

PR #125, opened 2017-09-13
Differentiated the buttons in Slim Mode progression so that button ty…

mtorchio:
…pes are clear. Status is a single selection so buttons are separated and have a border. Reason is a multi-select so buttons appear next to each other.

### DIFF
--- a/src/forms/ProgressionForm.css
+++ b/src/forms/ProgressionForm.css
@@ -1,6 +1,14 @@
 .button_disabled_is_selected:disabled {
     background-color: #297DA2 !important;
     color: #fff !important;
+    border: 2px solid;
+    border-radius: 10px; 
+}
+
+.button_disabled_is_selected:enabled {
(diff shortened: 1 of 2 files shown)


gregquinn2001:
did the mockup use the round cornered buttons vs rectangular ones as well?

Also, can we make the first button always align on the left side regardless of which type they are?

mtorchio:
Yes, the mockup used rounded buttons on top. I'm looking into the best way to align buttons to the left side.

gregquinn2001:
Maybe something like: https://css-tricks.com/snippets/css/remove-margins-first-element/

mtorchio:
We were setting the margins in code in the JSX style attribute, so I just made the change there. I tried what Greg's link recommended but couldn't get pseudo-selectors to move the buttons over.

mtorchio:
Thanks! I'm going to rebase and then merge this in.

PR #126, opened 2017-09-15
Progression changed to disease status

mtorchio:
I have changed everything user-facing in both FullApp and SlimApp. Variable names are unchanged and still contain the word progression.


gregquinn2001:
it's not directly the same, but the legend for the timeline view should be changed from "disease progression" to "disease status"

Dtphelan1:
Good catch @gregquinn2001!

Dtphelan1:
Made the suggested changes, merging!

PR #127, opened 2017-09-15
Reference date on progression slim

jafeltra:
This adds a reference date onto the progression form in slim mode. It uses a date picker, which I wasn't able to change the date format of, so I adjusted the format of the date in the copy button for the reference date and the as of date. The date that is saved in the Patient model is still the old format (ex. 15 Sep 2017). I didn't create a #reference date shortcut either yet, since that is a separate task I believe.

### DIFF
--- a/src/forms/ProgressionForm.css
+++ b/src/forms/ProgressionForm.css
@@ -65,3 +65,7 @@
 .helper-text {
     color: #b1b1b1;
 }
+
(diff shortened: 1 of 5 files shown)


gregquinn2001:
When I'm running it, I'm seeing a different date format for as of date than reference date (slim mode):

#progression  based on #Imaging relative to #reference date #05/21/2017  #as of #15 Sep 2017

PR #128, opened 2017-09-15
[Slim Mode] Improve sm screen display

Dtphelan1:
Various changes to improve the look of elements on the page when screen-width is restricted to half of a laptop/desktop monitor, including: 
- Font sizes are decreased slightly to make for less bloat without limiting readability.
- Prevent text overflowing of copy button when full on small screens.
- Slightly decrease the margins on small screens.
- Fix text-wrap on toxicity AdverseEvent dropdowns.
- Fix text overflow on toxicity Grade menu items when on small screens.
- Improved margins on disease status Status buttons to fix weird margins displayed when rows wrap.
- Increased margin-bottom of ShortcutViewer for better reading when copy button is full of content.
- Changed up helper-text padding for more uniform display on all screen sizes.

Let me know if there is anything else you flagged as looking odd on smalls screens and whether or not you would like me to make those fixes in this sprint or if we want to make separate tasks.

### DIFF
--- a/src/forms/ProgressionForm.jsx
+++ b/src/forms/ProgressionForm.jsx
@@ -71,6 +71,37 @@ class ProgressionForm extends Component {
         this.props.updateValue("referenceDate", formattedDate);
     }
 
+    renderStatusButtonGroup = (status, i) => { 
+        const marginSize = "10px";
+        const statusName = status.name;
+        const statusDescription = status.description;
+        const tooltipClass = (statusDescription.length > 100) ? "tooltiptext large" : "tooltiptext";
+
+        return (
+            <div key={statusName} className="tooltip">
+                <span id={statusName} className={tooltipClass}>{statusDescription}</span>
+                    <Button raised
+                    key={i}
+                    label={statusName}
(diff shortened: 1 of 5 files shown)


jafeltra:
This looks good! One thing that I noticed that I think related to the margin-bottom on ShortcutViewer. When the copy button has a lot of information in it, I can't scroll down enough to see the date picker for the Reference Date.

Dtphelan1:
Looks like this problem only comes up when the browser is restricted to mobile-phone sizes. Right now, for this task, I worked on improving styling for when browser width takes up ~1/2 of the screen.

jafeltra:
That make sense to me. This looks good then. Nice work!

Dtphelan1:
... Good enough to approve? :)

mtorchio:
This looks really good overall. One thing I noticed, and could probably wait for next sprint, is a decision on how the T, N, and M values should look. When those buttons become less wide than the text "#M0", it overflows. This may be the best way to handle this case, or perhaps we could hide the #?

Dtphelan1:
My bad -- the jira task was for slim mode changes. Changed the task headline.

PR #132, opened 2017-09-19
removed code block from the editor in patient mode

laclark22:
For patient mode there is no longer a block comment option on the tool bar.


gregquinn2001:
Looks good but I don't have a way to add a review so I'll do it in a comment! I'd say merge.

Dtphelan1:
Looks good to me! Does what you said, and it looks like you took care of the test file fix also!

mtorchio:
Looks good to me!

PR #133, opened 2017-09-19
improved scrollbar looks and darkened scrollbar for visbility

Dtphelan1:
Pretty self explanatory. Refers to JIRA task 534:  Scrolling is not always obvious to user; make changes to make scrolling more apparent to user.


nicoleng12:
Looks good!

PR #135, opened 2017-09-20
updated cheat sheets to PowerPoint format

mtorchio:


gregquinn2001:
I also e-mailed but thought it might be easier here:
GQ1	The Disease Status title is a smaller font than the Toxicity and Staging
GQ2	The Disease Status example should flow like regular sentences. There’s a period and the second sentence is incomplete.
GQ3	The label “Example:” doesn’t consistently align with the text to its right across the 3 slides

gregquinn2001:
It looks like we have the cheat sheets in 2 places now. Should the public folder only have the pdf versions that we serve up in the app and link to? Should the ppt versions be in the docs folders where the original ones were stored?

nicoleng12:
Looks good. I noticed the same things that Greg did. If you address GQ1, GQ2, GQ3, that will make the slides consistent with each other. 

In terms of file location, I agree with Greg. I think the public folder should be the document versions that the app uses whereas the docs folder is where the references are stored. Since the cheat sheet ppt isn't being used in the app, I think it should go in the docs folder

mtorchio:
I've replaced the old ones in docs/ with the up-to-date cheat sheets and addressed GQ1, GQ2, and GQ3.

gregquinn2001:
We can just have the pptx in the docs folder and the individual pdfs in the public folder. I think we can delete the combined pdf version from docs as it can always be created from the pptx if needed

mtorchio:
That's a good point Greg, I've removed the combined PDF.

Dtphelan1:
Are we trying to keep the descriptions we provide below for the structured phrases in sync with the descriptions provided in the cheat sheet? If so, we should decide whether or not we want to change the the descriptions in the cheatsheet to reflect the FluxNotes application's descriptions, or vice versa.  

Let me know if what I'm describing isn't clear.

mtorchio:
@Dtphelan1 are you referring to the differences between the text on the webpage and the text in the cheat sheet? All of it or a specific part? For example, there's a difference in the overall description of **Disease Status**:

Determination of disease status is based on a number of complex variables which include objective measures like tumor growth, symptomatic criteria, patient reports information, and subjective evaluations.

Based on the data available to the clinician at the time of evaluation, categorize the patient’s disease extent. Determination of disease progression is based on a number
of complex variables which include objective measures like tumor growth, symptomatic criteria, patient reported information, and subjective evaluations.

And a difference for **reference date**:

The date of the event that disease status is being assessed relative to. mm/dd/yyyy
 The #reference date is the date previous evaluations were performed, against which current evaluations are compared

Dtphelan1:
I was referring to the difference of overall description of Disease Status, in this case. Do we think that the description in the cheatsheet should reflect the fluxnotes description?

gregquinn2001:
@Dtphelan1 and @mtorchio  maybe we should close out this one as done and create a new jira to synchronize them and improve the documentation. we also may require deployment-/site-specific documentation; for example, for PATINA, the reference date is the date of randomization for the patient and we may not even have them enter it. that may be different for different clinical trials but also different for different sites within a clinical trial!

Dtphelan1:
I'd be fine with that!

mtorchio:
Sounds good to me @gregquinn2001 - you're right in that the wording might depend on the deployment.

gregquinn2001:
I approved it. Not sure if rebasing is required before merging

PR #136, opened 2017-09-20
Fix default reference date so 

nicoleng12:
Fixed the reference date in the disease status form in slim mode so that it doesn't default to today's date. By default it now shows "mm/dd/yyyy" until the user chooses a date.

- Set clinicallyRelevantDate to null in Patient.jsx when a new progression is created
- Edited progression form to use clinicallyRelevantDate for the default date (as opposed to a variable, like today's date, created in the form)
- Added code to convert date format to yyyy-mm-dd which is what the date picker component is expecting (didn't use moment.js to do this because kept getting console warnings that the code is deprecated)
- Added check for null value for clinicallyRelevantDate

### DIFF
--- a/src/forms/ProgressionForm.jsx
+++ b/src/forms/ProgressionForm.jsx
@@ -130,7 +130,29 @@ class ProgressionForm extends Component {
         )
     }
 
+    // formatDate converts javascript date format to YYYY-MM-DD
+    // @param date is a string of a date in D MMM YYYY format
+    // Source: https://stackoverflow.com/questions/23593052/format-javascript-date-to-yyyy-mm-dd
+    formatDate(date) {
+        var d = new Date(date),
+            month = '' + (d.getMonth() + 1),
+            day = '' + d.getDate(),
+            year = d.getFullYear();
+
+        if (month.length < 2) month = '0' + month;
+        if (day.length < 2) day = '0' + day;
+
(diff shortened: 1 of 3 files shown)


gregquinn2001:
I get the following if I haven't selected a reference date:
#disease status is #Responding relative to #reference date #12/31/1969  #as of #09/20/2017

nicoleng12:
The #as of #9/20/2017 is expected right? The concern is the 12/31/1969 date? That looks like Deborah's birthdate... I'll look into tracking down where it's pulling that from

mtorchio:
That's the start of the epoch, 0 seconds in Unix time, not her birthday (which was 1966 I think). It's probably trying to display the null value.

gregquinn2001:
yeah the reference date should be null and not included in the structured phrase until I provide one via the form, right?

jafeltra:
I think the changes you made look good! One small thing I noticed is that when you clear out the date with the x button in the picker, the date part of the string says #NaN/NaN/NaN. It might be coming from trying to reformat the date after it gets cleared out?

nicoleng12:
@jafeltra that is a really good catch. I added another check for the date in handleDateSelection. Let me know if this works now

jafeltra:
The extra check fixed it for me! Nice job! This looks good to me!

PR #137, opened 2017-09-21
Copy and paste shortcuts

gregquinn2001:
Copy and paste of shortcuts now works. Note that cut can be problematic just because deleting of multiple structured phrases at once doesn't always work even if it should (separate jira already to address although may be made a non-issue if we relax errors and support more flexibility in ordering of tags)

Went through definition of done and covered all items.

To test, you should be able to type a note including structured phrases and then copy it (control-C - need a jira to add to toolbar maybe?). then either move to end and paste or delete everything by starting at end and backspacing (see above about deleting multiple tags) and then paste.

The pasted content should match the original content (blue highlights for structured phrases) and the right panel should have contexts as appropriate.


nicoleng12:
I noticed an interested case where the pasting pastes more than what I highlighted. Here are the steps I did:

1. Click the @Name button in the context menu
2. Then in the editor after the name I typed "is diagnosed with " and then i click the @condition button and select invasive ductal carcinoma of the breast
3. Highlight "Deborah Hernandez", the structured phrase, and press control c 
4. Click the end of the note, right after "ductal carcinoma of the breast" and click control v

What happens is the entire note is pasted instead of just the name that I highlighted. Is it expected behavior that the entire note is copied? If it is then maybe we should indicate that users cannot select pieces of the note to copy?

gregquinn2001:
No that behavior sounds wrong. Let me take a look. Thanks!

gregquinn2001:
ok I fixed that and another problem I found. Definitely not confident but I couldn't break it with a number of different test situations I tried.

Dtphelan1:
If I type "blah blah @condition(select ductal)#staging", copy this, refresh the page, paste into editor, try to type #t1 or anything, the context panel on the right will display the cancer as the deepest level of context and will display #staging as a context preceding the cancer. 

Let me know if my description above wasn't specific enough to trigger this behavior on your end.

gregquinn2001:
Thanks @Dtphelan1! Good find. Let me see if I can reproduce and fix...

gregquinn2001:
Ok good catch @Dtphelan1. Thanks. I had to make sure context stack is built correctly even though when inserting a new shortcut there is no selection. Note parsing had a similar issue (no Slate at all!) so already had a solution available.

nicoleng12:
This looks good. I can select pieces of a note and paste it back in. I tried combinations of structured phrases with random text, just random text, and just structured phrases and it works. I also created a note with staging values, copied the T value, refreshed and pasted the copied structured phrase and correctly got an error message that the T value is invalid without the rest of the context. I also tested copying insertion of structured values from the summary panel via the plus button and that copies and pastes correctly as well.

Dtphelan1:
It looks like only content from the first line is being copied over onto the clipboard. For example, typing 'line one' on the first line and 'line two' on the second line, selecting all, copy, refresh and paste doesn't copy anything that wasn't on the first line. Interestingly, if I type 'aaaaaaaaaaaaaaa' on the first line, and 'b' on the second line, the clipboard will be populated with 'a'; it looks like the last line's length is being used to dictate how much content from the first line to copy. From my manual testing, the lines between the first and second don't seem to matter.

gregquinn2001:
Thanks @nicoleng12 . Appreciate the thorough testing. I'm feeling more confident now! :)

gregquinn2001:
thanks @Dtphelan1. I'll take a look. good test case

gregquinn2001:
@Dtphelan1 I can't reproduce. I typed "line 1" on the first line and then "line 2" on the second line and selected all then copied, refreshed page, then pasted and I got line 1 and line 2 as expected. Do you have the latest?

gregquinn2001:
oops. never mind. I was on master! sorry. let me try again.

gregquinn2001:
ok once I worked on the right branch, this wasn't too bad. We now support multi-line copies and pastes again.

jafeltra:
This looks good to me. I played around with different lines, different amounts of structured phrases being copied, inserting related phrases after pasting, refreshing and then pasting, and tried the bugs others had noted earlier. Everything seemed good to me and all of the previous bugs seem to have been fixed. This looks great to me!

jafeltra:
I haven't merged this in yet because I wanted to make sure that @Dtphelan1 had a chance to confirm that the bug he noticed is fixed now.

PR #138, opened 2017-09-22
Slim allow per site config

gregquinn2001:
View Manager now controls the list of shortcuts for a SlimApp
Apps now use the display value specified in ViewManager for each view/url
Moved from root URL to /patina
View Manager now supports configuration parameters that end up in form for the shortcut to support customization
Disease Status form supports configuration to show or not show reference date

### DIFF
--- a/launch_browser.cmd
+++ b/launch_browser.cmd
@@ -1,0 +1,2 @@
+start "C:\Program Files (x86)\Google\Chrome\Application\chrome.exe" http://localhost:3000/patina
+exit
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -16,9 +16,9 @@ class App extends Component {
                 <div>
                     {this.views.map((viewObject, i) => {
                         if (viewObject.isExact) { 
-                            return <Route exact path={viewObject.path} component={withTracker(viewObject.app)} key={i}/>
+                            return <Route exact path={viewObject.path} render={(props) => { return React.createElement(withTracker(viewObject.app), Object.assign(props, viewObject)); }} key={i}/>
                         } else { 
-                            return <Route path={viewObject.path} component={withTracker(viewObject.app)} key={i}/>
+                            return <Route path={viewObject.path} render={(props) => { return React.createElement(withTracker(viewObject.app), Object.assign(props, viewObject)); }} key={i}/>
                         }
                     })}
                 </div>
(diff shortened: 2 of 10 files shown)


gregquinn2001:
I guess go ahead and push as long as you tested the null check change. Were there cases where what was there wouldn't work though? The line above returns null or an object only. Sometimes the simpler straight-forward javascript check should be fine...

All new lines of code I wrote and any I modified should have used spaces instead of tabs so you must have gone ahead and changed other lines which is fine but doesn't really need to be part of this pull request...

Dtphelan1:
It 'works', but from a maintainability perspective there could be changes that we make in terms of code execution (e.g. having the return type go from null to '[]' or undefined that would go unnoticed if we didn't have an explicit check. When we have full-fledged testing in place, we should be okay; but our code should maintain those expectations when it comes to return types too. 

You're right, your new code is great when it comes to styling, but when we initially made the move from Tabs to Spaces, we talked about gradually changing the existing tabs to spaces as we went along and edited files. So I thought I would do that for this PR.

gregquinn2001:
There was an explicit check and it was based on the line right above it! I don't really agree that this one needs to be changed. The lodash methods more come into play when the subtleties of null vs. undefined and empty strings and all that can confuse things. when the line above sets something to null or an object instance then the next line checking if (ref) seems very straight forward and simple... just my opinion. we can discuss it more as a team at some point. thanks for the feedback.

Dtphelan1:
It's not that big of a deal, it was just a small suggestion. I won't push it, but will push the tab changes if that's okay.

gregquinn2001:
I agree it's not a big deal that's why I said in my first response go ahead and push if you want, but I do think it should result in a larger team discussion perhaps...

Dtphelan1:
But then you said you don't agree with the change, so I'll not include it for now and we can circle back to it if we want when we are having a larger discussion. I think there are some good questions to discuss as a team when it comes to best practices here, so that's a good idea!

nicoleng12:
Looks good! Navigating to http://localhost:3000/patina brings up slim mode. I tested by removing toxicity from shortcuts in the ViewManager and saw that slim mode removes toxicity from the list. Also played around with setting reference date to true and false to see it appear/not appear in disease status form.

gregquinn2001:
@nicoleng12 made a good suggestion to open browser pointing at patina url now instead of default url which is empty now. I committed changes to do that, but I doubt they will work on Mac. If someone with a Mac wants to try to find a solution that works there too, please do!

PR #139, opened 2017-09-25
Full mode as of date

jafeltra:
This creates a context for #as of and adds the shortcut #date as a child of that context. #date opens a calendar in the portal in the editor to pick a date. Let me know if this is the expected behavior.

One thing that I was unsure of and I'm open to feedback on is the decision to open a calendar in the portal. Right now, if an array of objects is returned by a shortcut, then the portal opens with a list of options instead of just inserting the text of the shortcut. In order to decide if that list should open a calendar instead of a list of options, I added a placeholder object that is structured the same as the list of options you would get with @condition. However, the portal will check if the key is equal to 'date-id' and if it is, it will open the calendar. This might not be the neatest way to do this, but was the simplest to work within the existing structure. If anyone thinks this should work differently, please just let me know.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "moment": "2.18.1",
     "positions": "1.6.1",
     "prop-types": "15.5.10",
+    "rc-calendar": "^9.0.4",
     "rc-tooltip": "^3.4.8",
     "react": "15.5.4",
     "react-autosuggest": "9.3.2",
--- a/src/context/ContextManager.jsx
+++ b/src/context/ContextManager.jsx
@@ -113,6 +113,15 @@ class ContextManager {
 	getPatientContext() {
 		return this.patientContext;
 	}
+    
+    getCurrentContext() {
+        const mostRecentContext = this.activeContexts[0];
+        if (!Lang.isUndefined(mostRecentContext) || !Lang.isNull(mostRecentContext)) {
+            return mostRecentContext;
+        } else {
+            return this.patientContext;
+        }
+    }
 }
 
 export default ContextManager;
--- a/src/context/ContextPortal.jsx
+++ b/src/context/ContextPortal.jsx
@@ -144,12 +144,11 @@ class ContextPortal extends React.Component {
     
     handleCalendarSelect = (date) => {
         this.closePortal();
-        const context = { key: 'set-date-id', context: `#${date.format("MM/DD/YYYY")}`, object: date }; //JULIA check about object
+        const context = { key: 'set-date-id', context: `#${date.format("MM/DD/YYYY")}`, object: date };
         const dateParentContext = this.props.contextManager.getCurrentContext();
-        dateParentContext.setAttributeValue("date", date.format("MM/DD/YYYY"), false);
+        dateParentContext.setAttributeValue("date", date.format("D MMM YYYY"), false);
         const state = this.props.onSelected(this.props.state, context);
         this.props.onChange(state);
-        
     }
     
     renderListOptions = () => {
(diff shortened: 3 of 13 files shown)


gregquinn2001:
[GQ1] When I hit Enter with the portal to select a date open, it closes the portal and puts "Placeholder for calendar" in the editor. It should go with the currently selected date or if not, do nothing I'd think.

jafeltra:
Oh I actually don't think I ever tried doing that. Good catch. I'll look at that now.

gregquinn2001:
overall this looks great though! Now we have the #date capability for other cases as well.

nicoleng12:
This looks good! It works really well. 

[NN1] - I noticed that when I copy and paste a note that includes the #date, when i paste, the date is not highlighted in blue as the other structured phrases are. 

For example, in patient mode, click on @condition and select "Invasive ductal carcinoma of breast". Select @disease status and then select @AS OF. Select #date and select a date from the calendar. In the notes section copy the whole line "Invasive ductal carcinoma of breast #disease status #as of #09/25/2017". Hit enter to go the next line and paste. All of the structured phrases are in blue except for the date

gregquinn2001:
Good test case @nicoleng12 . I can look into it if that would help @jafeltra . Let me know

nicoleng12:
[NN2] - In patina mode (slim mode), when reference date is set to false in viewermanger, the disease status form copy button still shows an as of date. I think a null check for the reference date should fix this issue

[NN3] - In patina mode (slim mode), when reference date is set to true in viewwermanager, the disease status form copy button shows 12/31/1969 as a default date. This should be null. I think this issue will get fixed when NN2 is dressed

gregquinn2001:
@jafeltra can you just log a jira for the copy and paste of #date values. I know what the problem is and fixing it is going to take some rework, but it's rework that should be done anyways IMO.

jafeltra:
Does NN1 have to do with the fact that you can't type #09/25/2017 to insert a date but instead have to type #date and then pick from the calendar? I haven't looked at it, but that's my original guess. I'm still working on the fix to GQ1, so @gregquinn2001 if you want to look at it, feel free! I can also just log a jira for it and rework it later.

gregquinn2001:
Actually, no. Supporting typing a date directly in after a hashtag would be new capability unlike anything we have so far. You can't just type the condition you want either right now. Also, though, condition is an inserter and date is a creator.

gregquinn2001:
ok. I fixed NN1. I still want to revisit how we are identifying the type of data a shortcut represents as Julia describes in her pull request, but this solves the problem.

gregquinn2001:
Just need to fix GQ1 and NN2 now and then I'll approve this. Thanks!

gregquinn2001:
[GQ2] Just realized that #as of is not disappearing from context after it's entered into note. this was that strange case I think we discussed. I'll take a look at fixing this one and commiting to this branch..

gregquinn2001:
GQ2 fixed now. #as of is removed from right once added to editor and re-appears if #as of is deleted. Copy and paste seems to end up with the right context state

nicoleng12:
I believe the edit i made to the check for undefined (I instead check for existence so it covers the null case) fixes NN2, but i used LastUpdateDate instead of OriginalCreationDate because LastUpdateDate contains today's date. I think we should discuss which date we should be using for the "as of" date

gregquinn2001:
as of date should be OriginalCreationDate as it is being created for this. I guess maybe when created, we should consider whether LastUpdateDate should be set as well. Not sure if SHR specifies this, but it should so that implementors are consistent.

jafeltra:
I just pushed up changes, that should fix GQ1, update's Nicole's fix for NN2, and updates how the date's attribute value is set so that it works correctly within our structure. If anyone can look at the changes made and let me know if they find anything else that should be fixed, please just let me know!

nicoleng12:
This looks good! GQ1, NN1, and NN2 have been fixed.

PR #141, opened 2017-09-28
Reference date to disease status

mtorchio:


jafeltra:
This looks good to me! I think that you need to add the same check for "referenceDate" to not be true in shouldBeInContext() as you added in getValidChildShortcuts(). That lets the #disease status context stay open if you've entered everything but the reference date. I think that is what we want, but I'm not totally sure. That's at least what as of date is doing now.

jafeltra:
One other thing that I found is changing the reference date on slim mode. I think since you updated the case in setAttributeValue in ProgressionCreator to "referenceDateDate", when you pick from the datepicker, you need to use that string instead of "referenceDate". This is also really minor, but I don't think this.resetReferenceDate is actually needed anymore since Nicole updated it to handle starting out as null. I just noticed this now though so you don't have to change it in this pull request.

gregquinn2001:
@jafeltra In ProgressionCreator. Yes this is true. Everywhere "asOf" appears should be copied and replaced with "referenceDate" in the copy! Good catch.

gregquinn2001:
@jafeltra . I think we should do both of these before merging. Julia, if I get these wrong please let me know.

JA1 - Update ProgressionCreator.shouldBeInContext to check for referenceDate to match what it does for asOf
JA2 - Update ProgressionForm.handleDateSelection should use "referenceDateDate" as the first argument.

jafeltra:
@gregquinn2001  and @mtorchio  Those look right to me! The only other thing I mentioned was:

JA3 - this.resetReferenceDate in ProgressionCreator is probably not needed any more.

Since this is related to old changes, I'm fine if we don't include that in this pull request.

gregquinn2001:
I think we should just do all 3 of Julia's items. Mike, they all seem very quick. You'll spend most of your time re-testing. If any of them are not quick, please ask Julia or myself for some assistance.

mtorchio:
Okay, I pushed a commit to resolve JA1, JA2, and JA3.

jafeltra:
They all look good to me!

PR #142, opened 2017-09-29
Enable toolbar buttons

laclark22:
The toolbar is now active. When you click on one of the buttons it will bold the symbol and as you type in the editor the text with have the right style. Once you delete all text with that specified style in the editor the button will gray out again. For bulleted-list and numbered-list only the button can toggle them on and off once you backspace to the initial bullet/number. I have added a issue in JIRA for this since I wanted to get this pull request in before the end of the sprint and I believe that might take a little more investigation since the current slate demo for the toolbar also has this problem (http://slatejs.org/#/rich-text)

### DIFF
--- a/src/notes/EditorToolbar.css
+++ b/src/notes/EditorToolbar.css
@@ -38,7 +38,7 @@
 }
 
 .MyEditor-root .autocomplete-menu .menu-item[data-active="true"] {
-    background-color: rgba(0,0,0,0.1);
+    background-color: rgba(211,211,211,0.1);
 }
 
 .MyEditor-root p {
--- a/src/notes/EditorToolbar.css
+++ b/src/notes/EditorToolbar.css
@@ -38,7 +38,7 @@
 }
 
 .MyEditor-root .autocomplete-menu .menu-item[data-active="true"] {
-    background-color: rgba(211,211,211,0.1);
+    background-color: rgba(0,0,0,0.1);
 }
 
 .MyEditor-root p {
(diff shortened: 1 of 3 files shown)


Dtphelan1:
Looks good! 

[DP1] - Non-list-items written after the lists require two enter presses instead of one to make a new line. This has to do with the value for DEFAULT_NODE being set to 'paragraph'. If we change the default node type to be an empty string instead, new text gets inserted as a Div instead of a Paragraph, and the double enter press bug goes away. 

To trigger, type: 
Sample Text: 
- This is a bulleted list 
More Sample Text 
Requires two enterkey presses 
to go to the next line.

jafeltra:
This looks good to me! One other thing I noticed:

JA1: If you have text bolded, then insert a shortcut, the text before the shortcut becomes unbolded.

nicoleng12:
Nice job! It looks really good.

I noticed a strange error:

[NN1] - If you insert a shortcut and then add another shortcut, set the cursor to right at the start of the second shortcut and click enter,  the second shortcut gets copied and pasted into the next line

Steps to reproduce:
1. In patient mode click @name then click @age 
2. Set the cursor to right at the start of @age 
3. click enter. With each enter press the shortcut gets copied again

image

nicoleng12:
[NN1] seems to be an issue in master too... I'll make a jira for it. I don't think you need to address [NN1] in this branch

gregquinn2001:
What does everyone think then? JA1 and DP1 seem to be the ones that should resolved prior to merging?

gregquinn2001:
NN1 is pretty interesting but I'm glad you tested on master too!

Dtphelan1:
I think DP1 should definitely be fixed, it's a single change I've already ID'ed. I could go either way on JA1 -- that is, fixing it now or lodging a bug.

laclark22:
Addressed DP1 and made a JIRA for JA1

nicoleng12:
Looks good! Dylan's error has been addressed. I'm going to merge this

PR #145, opened 2017-10-02
Upgraded to material-beta-13

Dtphelan1:
Upgraded to beta 13, changed one thing in Nav, didn't see any other issues. Someone wanna check their machine for  full functionality too?

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "font-awesome": "4.7.0",
     "guid": "0.0.12",
     "immutable": "3.8.1",
-    "material-ui": "1.0.0-beta.5",
+    "material-ui": "1.0.0-beta.13",
     "material-ui-icons": "1.0.0-beta.5",
     "mkdirp": "0.5.1",
     "moment": "2.18.1",
--- a/src/nav/NavBar.jsx
+++ b/src/nav/NavBar.jsx
@@ -8,8 +8,7 @@ import Typography from 'material-ui/Typography';
 import Button from 'material-ui/Button';
 import IconButton from 'material-ui/IconButton';
 import MenuIcon from 'material-ui-icons/Menu';
-import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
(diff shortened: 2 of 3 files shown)


mtorchio:
Looks good to me. The only item I noticed was a new warning: 

Warning: Material-UI: the value provided `4` is invalid

nicoleng12:
Everything looks good! I tested both in chrome and in IE as well as ran the unit tests. All the unit tests passed. 

I also see the warning that Torch saw. As he wrote, it happens when you click on the calendar. If the fix is not super straightforward, should we close this request, merge it, and make a jira for the material ui warning?

gregquinn2001:
I would say it makes sense to look into and perhaps understand what's causing the warning. If knowing that, it doesn't affect us, then we can merge this and log a jira to eliminate the warning. If the exploration is taking too long, record what you learned in the JIRA and close it since the task was to explore and only implement if no work was required. Make sense?

Dtphelan1:
That's weird. I'm not getting that warning... When one of you is in the office, I'd like to come over and take a peek! Let me know.

laclark22:
I don't see any warnings when I click on the calendar. Very strange that some of us do and some of us don't...

jafeltra:
So I realized this morning that I forgot to run yarn install, so my testing of this was completely useless and I'm getting the same warning now, and also an error "Uncaught RangeError: Maximum call stack size exceeded". Sorry for not thinking yesterday!

Dtphelan1:
I'm now able to get the bug. It seems like the update has somehow caused a lag in setState for the ContextTray's Tabs. This is a breaking bug, and because of it I don't think we should update our beta. We can revisit the Material ui beta update when we change our Context Tray's UI away from the Tabs and towards the sectioned buttons, but for now it looks like we shouldn't put any more time into this.

gregquinn2001:
Sounds good. Please record what you learned on the JIRA (make sure it says the specific version you were going from and to). You can reference this pull request thread directly from a comment.

Dtphelan1:
Can do!

PR #146, opened 2017-10-03
Remove ripple from copy button (couldn't change ripple color). 

Dtphelan1:
Shocker: Turns out changing the color of the ripples is, as far as I can tell, obscured away from us. I tried playing with MUIThemeProvider and specifying ripple styling – no luck. I tried passing a prop called rippleColor with new value – no luck. (N.B. these were the top stack overflow suggestions I could find).

One alternative solution would be to disable the ripples on the copy button altogether, and blur focus from the button onClick. I've implemented this and have it on a branch, but we should review it as a team and decide if this is something we want. Furthermore, if someone can see how to change ripple color, we should implement that and consider it as an option -- I wasn't able to in the timebox I allotted.

### DIFF
--- a/src/viewer/ShortcutViewer.jsx
+++ b/src/viewer/ShortcutViewer.jsx
@@ -1,4 +1,5 @@
 import React, {Component} from 'react';
+import ReactDOM from 'react-dom';
 import CopyToClipboard from 'react-copy-to-clipboard';
 import Button from 'material-ui/Button';
 import LandingPageForm from '../forms/LandingPageForm';
@@ -30,6 +31,7 @@ class ShortcutViewer extends Component {
                     <Button 
                         raised 
                         className="btn_copy"
+                        ref="copyButton"
                         style={{
                             textTransform: "none",
                             justifyContent: 'left',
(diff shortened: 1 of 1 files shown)


gregquinn2001:
Maybe this is a dumb question but if we removed the focus ripple why do we need to blur on click as well?

Dtphelan1:
When you have focus on the button it looks lifted off the page (longer and darker box shadows) emphasizing that it's selected. When you blur it provides (granted, a modest, but albeit) a visual indicator that something has happened.

Dtphelan1:
At least that was my justification*

nicoleng12:
In order to test this, we click on the copy button, alt tab out of the app then alt tab back in right?

Since I was not able to reproduce the original ripple effect error, I am also not seeing the latest fixes...

nicoleng12:
After talking with Dylan, the effect is triggered from tabbing between buttons as well (not just alt tab). When I click on a button and then tab to bring focus to the copy button, I see that the button has a blur indicating that an action has occurred on it. The change is subtle but enough to indicate that the copy button has focus. Looks good to me!

PR #147, opened 2017-10-05
Changed Timeline to use asOfDate

danlee1025:
Only a couple lines changed(if you do a git diff -w), but I refactored this function to follow the style guide.

### DIFF
--- a/src/timeline/TimelinePanel.jsx
+++ b/src/timeline/TimelinePanel.jsx
@@ -289,35 +289,35 @@ class TimelinePanel extends Component {
   }
   
   _createProgressionItems(patient, progressions, groupStartIndex) {
-    let items = [];
+      let items = [];
+      
+      progressions.forEach((prog) => {
+          const assignedGroup = this._assignItemToGroup(items, prog.clinicallyRelevantTime, groupStartIndex);
 
-    progressions.forEach((prog) => {
-		const assignedGroup = this._assignItemToGroup(items, prog.clinicallyRelevantTime, groupStartIndex);
+          let classes = 'progression-item';
+          // Do not include progression on timeline if asOfDate is null
+          if (prog.asOfDate == null) return;
+          let startDate = new moment(prog.asOfDate, "D MMM YYYY");
+          let hoverText = `${startDate.format('MM/DD/YYYY')}`;
+          let endDate = startDate.clone().add(1, 'day');
+          classes += ' point-in-time';
+        	
+          let focalCondition = patient.getFocalConditionForProgression(prog);
+          let focalConditionName = focalCondition.specificType.coding.displayText;
+          
+          let hoverTitle = focalConditionName + " is " + prog.value.coding.displayText + " based on " + prog.evidence.map(function(ev){ return ev.coding.displayText; }).join();;
 
-		let classes = 'progression-item';
-		let startDate = new moment(prog.clinicallyRelevantTime, "D MMM YYYY");
-
-		let hoverText = `${startDate.format('MM/DD/YYYY')}`;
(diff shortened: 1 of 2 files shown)


jafeltra:
The changes look good to me! The icon and hover text use the as of date now. I didn't notice the warnings that Mike saw. I like the idea of pointing out the updates to the timeline in the demo script. I also agree that it would be useful to start adding a few tests around new features so we can start to keep up with the changes we make.

gregquinn2001:
@Dtphelan1 could we do a UI test that would create a new progression by entering #disease status #stable ... into the editor and then verifying that the summary and timeline are updated appropriately?

gregquinn2001:
I already logged bug ASCODCP-596 (https://jira.mitre.org/secure/RapidBoard.jspa?rapidView=553&projectKey=ASCODCP&view=planning&selectedIssue=ASCODCP-596) because I don't think the summary is updating correctly so it would be ok if that part of the test was left to that bug or if it just failed for now...

Dtphelan1:
Yeah @gregquinn2001 that's exactly what I was thinking. @daniellee1025  want to put something together that meets those testing requirements? For information about the different types of testing libraries we use look [here](https://github.com/standardhealth/flux/wiki/Flux-Capture-Prototype-%28Flux-Notes%29#technologies-used) (you'll probably want to look into the testcafe library). And for examples of the testing, you can check the test/ui/ directory in the project. Let me know if you need any help with this!

danlee1025:
@Dtphelan1 I will look into and ask you for help if needed.  I'm mostly working on my other project today so I probably won't get to it until Monday.

gregquinn2001:
The test fails when I run it "Typing a progression note with as of date in the editor results in a new progression item on the timeline"

danlee1025:
@gregquinn2001 I'll have the fixed test pushed up soon.  I was pushing up the inital test to work with Dylan.

gregquinn2001:
Sorry Dan!

nicoleng12:
This looks good Dan. The timeline is using as of dates and when I add a disease status with an as of date, it gets added to the timeline. I think once the tests are added and they work/pass, this branch should be good to merge with master

danlee1025:
@gregquinn2001 @nicoleng12 I pushed the UI test up now and all the tests pass on my end.  Can you guys take another look?

danlee1025:
Should remain consistent if patient data does not change.  I tried looping through the progression items and checking if any of the progression items contained today's date and setting a boolean to true and then asserting after the for loop.

I could not get this condition `if(hoverTextItemDate.innerText === today.toString())` to work, but the assert equality works.  Any ideas?

jafeltra:
One thing that I ran into when trying to use the `.innerText`, is that it seems to be asynchronous. If you add `await` before getting the `.innerText` it might help. Something like: `const text = await hoverTextItemDate.innerText` and using that `text` variable in the comparison. Maybe that would help here too?

nicoleng12:
Julia and I worked on some of the unit tests together and I just want to second her comment. Adding an await should help you set your variable because it's asynchronous. I had to do that in my unit tests as well.

Also, all your tests pass. I can approve this and merge but @danlee1025 are you trying Julia's suggestions first? If you are, I can hold off

danlee1025:
@nicoleng12 I would like to try out Julia's suggestion, will be pretty quick.

danlee1025:
@jafeltra @nicoleng12 Thanks for the help, that worked.  I had an await when selecting the element so didn't even think to put an await when selecting the innertext.  It should be good to go now.

nicoleng12:
All the tests pass. Nice job Dan! I'm merging this

PR #148, opened 2017-10-06
Shortcut refactoring + more (554, 519, 526, 555, 595)

gregquinn2001:
[JIRA 554] #12/22/2014 being entered into editor is now supported. #date still works to pop up date selector as well. Only #date appears in right panel. Now infrastructure supports regular expression specified value shortcuts through new getTriggerRegExp. A shortcut can provide both string triggers (#date) and a regular expression trigger (actual date). I added a UI test case for entering an actual date into the editor.

[JIRA 519,526] Parsing of notes containing structured dates is now supported. Parser was refactored to support the string triggers and regular expressions as well as provided by shortcuts. Test cases were added for parsing notes with dates.

[JIRA 555] Parsing now also detects unexpected # or @ expressions and returns a list of them from the parse method in addition to the SHR patient. An existing test case already tested for this already so once all test cases were updated for parser API changing to return both the SHR entries and the unexpected structured phrases, that test case passed and verifies that the unexpected tag is returned.

[JIRA 575, 595] Pasting of plain text structured phrases now works as well; however, @condition and #date cause issues due to the portal pop-up. @condition[[Lobular Carcinoma...]] can be used but I left the issue open for this so we can revisit how those are handled. I think 595 can be marked fixed but 575 should stay open. Also, test cases need to be added for pasting which can be done as part of 575.

### DIFF
--- a/src/context/ContextOptions.jsx
+++ b/src/context/ContextOptions.jsx
@@ -30,7 +30,7 @@ export default class ContextOptions extends Component {
         // count how many triggers we have
         let count = 0;
         validShortcuts.forEach((shortcut, i) => {
-            shortcut.getTriggers(context).forEach((trigger, j) => {
+            shortcut.getStringTriggers(context).forEach((trigger, j) => {
                 count++;
             });
         });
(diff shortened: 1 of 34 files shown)


gregquinn2001:
Please log a detailed jira bug Dylan. good one. I think this is a general problem? In other words, this problem is not specific to dates?

Dtphelan1:
Thanks! Yeah I think a separate JIRA will be good. It's a general problem with how we've approached regex's for AutoReplace

jafeltra:
This looks really good to me! One thing that I noticed is:
[JA1]:  If you copy and paste a date structured phrase, the date is pasted twice. 

I'm not sure if this is related to some of the pasting issue that Greg suggested remain open. If so, this might be fine to hold off on. 

Besides that, all the test cases passed, the parsing seems to be working correctly, and I can enter in dates using #10/10/2017 and they becomes structured phrases and update the patient information.

gregquinn2001:
Thanks for reviewing Julia. I'll take a look at JA1 before we merge.

gregquinn2001:
JA1 fixed and all tests still pass. It would be nice to add some copy and paste test cases so maybe I'll try to do that. I'll need to get a lot better at testcafe!

gregquinn2001:
I'd vote to merge it now.

jafeltra:
Sounds good! Merging now!

PR #149, opened 2017-10-10
Deceased shortcut

nicoleng12:
Created deceased shortcut so users can now add date of death information for the patient.

**Shortcut:**
- In slim mode, deceased shows up as an option in the list on the left . Clicking on "deceased" in the list brings up the deceased form. Users can use the date picker to select the date of death. The copy button updates with the selected information and users can click on the copy button to copy the deceased information
- In full mode, #DECEASED buttons shows up in the patient context panel. Clicking on the button inserts the deceased shortcut and users are then presented with the #DATE button in the context panel. Clicking on the #DATE button brings up a date picker for users to select the date of death. Users can type in the shortcut (both 'deceased' and the actual date). Users can copy and paste the shortcut. Functionality works the same as existing shortcuts

**Cheat Sheet:**
- "Structured Phrase Format Descriptions" Powerpoint in /docs has been updated to included deceased
- In slim mode, on the deceased form, users can click on the link which displays the cheat sheet for deceased as a pdf

**Tests:**
- Existing tests passed 
- Added UI tests for slim mode and full mode
- Added backend tests
- Edited note parser test to include testing of deceased sample note

**Note Parser:**
- Updated note parser to include a sample note for decaesed
- Ran parser and it successfully parses the deceased note and creates an object in the expected format with the expected values


gregquinn2001:
I think the addEntryToPersonOfRecord needs to be renamed we don't add entries to the Person of record - only to the patient's SHR that method should be setDeceased probably and the argument is a deceased object so let's rename the argument to deceased.

gregquinn2001:
Also, in the DateOfDeathCreator, I have a minor nit. The getDateOfDeathString should not return the #date with a space before it. The space should be added when it's combining the date of death string with the #deceased in getAsString IMO.

gregquinn2001:
[GQ3] In ViewManager, reference date got turned on for disease status. It should be false for PATINA.

nicoleng12:
Great feedback! I will work on addressing GQ1, GQ2, and GQ3

gregquinn2001:
[GQ4] Update the description of deceased appearing on the form in slim mode to eliminate the talk about the data. Maybe change it to something like "An indication that the person is no longer living as of the date specified below."

nicoleng12:
I've addressed issues GQ1, GQ2, GQ3, and GQ4

jafeltra:
I think this looks really good! The short cut is added correctly, can be deleted, and copy and pasted. The slim mode form works as expected. All the tests pass as well, and the new note is parsed correctly.

One question I have is should the #deceased button in the context panel be removed once #deceased is added to the note? I had this thought, but then realized that we aren't removing shortcuts like #staging, so maybe it should stay as it is to keep things consistent. @gregquinn2001  do you know which way is best?

gregquinn2001:
@jafeltra . good question. staging, progression, and toxicity can all have multiple instances per condition. deceased though should only happen once per patient really. I do think it should not be shown if the patient is already deceased

nicoleng12:
I can hide the button by doing a check in patient context to see if there is already a deceased object. If there isn't, then the deceased button will be displayed, and if there is a deceased object, the button will not show up. 

But, what should happen when the user deletes #deceased #<date> from the notes? Because the patient is updated with having a deceased object and we don't currently have functionality to remove data from the patient when we delete shortcuts from the notes, the button will never show up again unless you do a refresh. 

I can:

a) Make a jira issue for hiding the deceased button and merge as is

b) Incorporate the changes that removes the button and make a jira issue for getting the button back when we delete the shortcut (if that's the behavior we expect)

c) If anyone has an idea for how to handle it and it's a small quick fix I can incorporate it into this pull request

jafeltra:
I would vote for option a. I think that since it seems like it would be more work to correctly delete the deceased shortcut (and other shortcuts) and update the visibility of the button, it should be done separately. And I think having the button there is better than not being able to get it back at all once it's deleted.

gregquinn2001:
You need to check if data set on patient in the shortcut's validateInCurrentContext and in validInContext methods depending on which shortcut. We can discuss in person tomorrow.

The onBeforeDeleted method should call setAttributeValue to the value it represents to null so it shows up as not set and the trigger appears in the right panel again.

nicoleng12:
Sounds good let's discuss tomorrow how to go about doing this.

I rebased off of master and some of UI tests are broken now... working on fixing them at the moment

nicoleng12:
Thanks for the fix Dan! Tests are passing now. I'm looking into hiding the deceased button

nicoleng12:
Rebased and all the tests passed

PR #150, opened 2017-10-10
Patient enrollment shortcut

jafeltra:
This creates a clinical trial enrollment shortcut so users can now add a study enrollment to the patient's record. The clinical trial shortcut allows users to specify the title of the clinical trial, the enrollment date, and an end date of the trial.

**Shortcut:**

- Slim mode: Clinical trial is now an option in the list on the left. Users can select the title of a clinical trial and optionally select an enrollment date and end date using the date picker. The copy button updates, only adding dates once a trial title is selected.

- Full mode: a #clinical trial button is available at the patient level in the context panel. This is because the clinical trial is on a patient, not a specific condition like our other shortcuts. The button does not disappear after added to a note since a patient can be in multiple clinical trials. Once #clinical trial is inserted, it has child structured phrases for title, enrollment date, and end date. The title structured phrase allows a user to type or click on a specific clinical trial title (ex. #PATINA) or type or click the general #title shortcut, which opens a portal with the list of possible trials. Enrollment date and end date work the same as the dates in disease status. You must first specify #enrolled on or #ended on and then choose the date using #date.

**Cheat Sheet:**
- The powerpoint file in /docs was updated to include a slide for clinical trial.
- A pdf version of the new slide was included in /public and is linked to in the clinical trial form in slim mode.

**Tests:**
- Existing tests all pass.
- Added UI tests for slim mode and full mode.
- Added backend tests for Patient.jsx to ensure clinical trial enrollment is correctly added to the patient record and updated.
- Added a note parser test to test parsing a note with #clinical trial.

**Note parser:**
- Added a sample note that includes #clinical trial structured phrase. The parser correctly parses this note.

All feedback on this is greatly appreciated!


gregquinn2001:
[GQ1] ClinicalTrialsList.jsx doesn't use the class format like our other data. Is there a compelling benefit to using the format used? Should we switch our other data classes? We should be consistent so either we need a JIRA to switch Patient to the non-class format or we should make ClinicalTrialsList a class.

gregquinn2001:
Should we create a JIRA to create a common date picker component to use on our forms? I see the format appearing in a lot of places in our code for example, so it would be nice to control that in 1 place.

gregquinn2001:
[GQ2] This is my fault I think, but StudyEnrollment should be added to the patient SHR using addEntryToPatientWithPatientFocalSubject. Keep the addEntryToPatient method in Patient though.

gregquinn2001:
[GQ3] We should have a sample note with the minimum required subtags for the clinical trial so we can make sure the default values look right.

gregquinn2001:
[GQ4] The description for #title should not be so generic since the tag is specific to the title of a clinical trial. Also, should it be #title or #name? I think they typically refer to the name of a clinical trial?

gregquinn2001:
Should the end date on the form have a clear way of specifying "no end date" because the patient is still enrolled? Maybe a checkbox labelled "Clinical Trial involvement has ended" before the end date that defaults to off and disables the date picker until it is checked? Thoughts?

mtorchio:
I think a common date picker would make sense rather than duplicating things, I vote to create a JIRA issue for that.

gregquinn2001:
Thanks @mtorchio. I'll wait to see if there's any other thoughts on it and if there is consensus I'll create the JIRA.

jafeltra:
I agree that a separate component for the date picker would make sense, so I would vote for a separate JIRA issue for that. I will make changes to address GQ1, GQ2, and GQ3. I will also add some helper text to indicate that the end date is optional, and we can evaluate the design of the form with the design team. For GQ4, I named it #title based on the SHR Study data element, which has a title and an identifier attribute. If we decide that the shortcut should use #name instead, I think we can adjust that fairly easily. I can update the description for #title to be less generic though.

nicoleng12:
I agree that we should have a separate JIRA issue to make a separate component for the date picker since it gets used a lot in our code

gregquinn2001:
added JIRA ASCODCP-602

jafeltra:
I just updated the pull request to fix GQ1, GQ2, and GQ3. For GQ4, I changed the description of #title, but kept it as title instead of name to stay consistent with the SHR element. I also added helper text to the End Date on the Clinical Trial Form to indicate that it is optional.

gregquinn2001:
noticed one test (update methods for study enrollment) still uses addEntryToPatient instead of one with focal subject

gregquinn2001:
For GQ3, I meant to add another sample note with minimum elements for clinical trial tag to the note parser test. I wasn't clear. Did you interpret this differently?

jafeltra:
Oh. Yes I did. That was my fault. I added another sample note with just #clinical trial. I can do the same in the noteparser.test file.

jafeltra:
Whoops. Good catch on the `addEntryToPatient`. I missed that one. I'll fix that now with the notepareser test.

laclark22:
Looks great just make sure you rebase before anyone merges this

jafeltra:
Thanks for the approvals! I'm going to work on rebasing this now. I can merge once it is set.

jafeltra:
I just rebased and all tests are passing. I'm going to merge this now.

PR #151, opened 2017-10-11
Data access abstraction layer (JIRA 584)

gregquinn2001:
went through definition of done

Added DataAccess class used by NoteParser and FullApp to abstract how Patients are loaded, saved, and created. Uses a configured data source behind scenes of DataAccess. 2 data sources behind the DataAccess class are supported
(1) hard coded read only patient access supports the hard coded patient only in read only mode.
(2) new patient only for note parser to use only allows creation of new patients.

FullApp supports configuration per url now (via ViewManager)  that can specify name of data source to use. It instantiates DataAccess providing desired data source and then uses that to get the demo patient (demo patient's ID defined within DataAccess)

Ready to support a data source that is a client to our planned patient management service.

Tests added for all methods of both supported data sources

Data Sources will need to be changed to leverage new object model once that's ready.


Dtphelan1:
[DP1] HardCodedReadOnlyDataSource doesn't have a constructor that creates a single hard-coded patient/list of patients. This results in a new list of patients being made on every call to getListOfPatients. 

I've written a test that checks to see that the lists of patients returned on consecutive calls of getListOfPatients are strictly equal to one another -- let me know if you want me to push this.

Dtphelan1:
[DP2?] /test/ui/FullApp.js loads Patient directly, instead of loading it through the data access interface. (I put a question mark because this might be intentional -- just wondering what the justification for that is.)

gregquinn2001:
On DP1, no I don't. This is a temporary class or test-only class at most going forward. It's fine as it is.
On DP2, it's fine for a test case to load it's reference data in a way that minimizes possible bugs (i.e., as directly as possible). I prefer it to load it directly rather than go through the data access layer because it's not trying to test the data access layer.

gregquinn2001:
Do people agree with that logic for DP1 and DP2? That was my thought process, but I meant to access what everyone thinks. I typed my response quickly since I was late for stand up!

Dtphelan1:
I'm wholly convinced on DP2 (for the record) but I'd like to hear what people's thoughts are on DP1. I'm fine without the constructor, but it feels odd that we're making a new object using the same JSON each time we call getPatient or getPatientList. I get that this isn't a permanent solution though.

gregquinn2001:
For a test class I usually try to keep things real simple and performance is not a concern unless it becomes an issue.

Dtphelan1:
That makes sense to me -- making extraneous tests for scaffolding is just extra work now and extra work to undo when the full implementation is done! I'm comfortable deferring to your judgment here -- plus, you've got considerably more experience than I do in both class-driven software dev and in constructing full REST-based services like we're doing for this task.

jafeltra:
This looks good to me too. I think Greg's reasoning makes sense for DP1 since eventually the new hardcoded patient will be replaced. I ran all the tests and everything passed, the noteparser correctly parses all notes, and I couldn't break anything in the app.

nicoleng12:
This looks good. I went through the code and ran the tests. All tests pass (except for one that I always seem to have trouble with, but I think that's just on my end and not related to this branch). The note parser works as well. Since we are eventually redoing new hardcoded patient, then I think DP1 will not be an issue later on. Reasoning for DP2 makes sense. If everyone is good with this I shall merge

PR #178, opened 2017-11-09
unifed docs formatting, font sizing, and label locations

Dtphelan1:
Removing the typical PR template because of the limited nature of these changes. Just updated the cheat sheet documents to unify the styling a little bit. Take a look at let me know what you think, if there's anything you would add, etc...


jafeltra:
This looks good to me! The only thing I'd add is to update the pdf versions of each slide so we link to the updated slides in slim mode.

Dtphelan1:
Added new single sheets based on Greg's changes

PR #184, opened 2017-11-16
Implemented New Design for Landing Page

noranda:
# Implemented New Design for Landing Page

* Added the following libraries:
   * `node-sass-chokidar 0.0.3` - Added **SASS** for CSS pre-processing. Allows use of shared variables, mixins, nested css, partials, etc.
_Note: When working with Sass files (`*.scss`), run `yarn run watch-css` to process CSS in real-time._
   * `react-fontawesome ^1.6.1` - Added **Font Awesome** react components to provide simple API for using font awesome icons including modifiers.
* Added Open Sans font via Google fonts including all variations, which was not correctly imported before now.
* Added `png` and `svg` landing page assets in `public/landing`.
* Updated `public/logos` to transparent versions.
* Implemented new design for Landing Page in `src/apps/LandingPage.jsx` including:
   * Responsiveness from desktop (Chrome & IE) to mobile (Chrome).
   * Links to full and lite version.
   * Included icon svg's (instead of directly sourcing) for use in IE.
* Moved landing page tests to its own file under `/test/Landing.js` and added test for all links.

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- [x] **N/A** Cheat sheet is updated
- [x] **N/A** Demo script is updated 
- [x] **N/A** Documentation on Wiki has been updated 
- [x] Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- [x] **N/A** Note parser has been updated if structured phrases change

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed
- [x] **N/A** Added UI tests for slim mode 
- [x] **N/A** Added UI tests for full mode
- [x] Added UI tests for Landing Page
- [x] **N/A** Added backend tests for fluxNotes code
- [x] **N/A**Added note parser examples to test new functionality

## Reviewer 1: Name

## Reviewer 2: Name


gregquinn2001:
[GQ1] Flux Notes Lite and Full app bar banner at top is now centered instead of left-justified per designs.

gregquinn2001:
[GQ2] Typo on landing page: bottom right: "approch"
[GQ3] 2nd to last paragraph on bottom right. Should "condition, domain, task" each have a hyphen after them as they really saying condition-specific, domain-specific, task-specific, and role-specific?

gregquinn2001:
[GQ4] The initiatives at the bottom of landing page aren't supposed to be links?

BTW, the page looks awesome overall. I like the screen shots for linking to the prototypes!

jafeltra:
This looks really good! Nice job! A few things I noticed are mostly related to styling on the other pages besides the landing page. 

[JA1] There seem to be more scroll bars in both full and slim mode.

[JA2] There is a white boarder on the left side of slim mode that wasn't there before.

[JA3] The text is not centered vertically in the copy button in slim.

[JA4] The scroll bar for the slim mode page is behind the copy button. Before we had purposely shortened the width of the copy button so we could see the full scroll bar.

[JA5] I'm not sure if this is something that needs to be changed, but I noticed it as different so I thought I'd note it. The headings for the individual child shortcuts to add in slim mode (like Adverse Event, Grade, Attribution) are bigger than the main page title now (Toxicity).

Dtphelan1:
[DP1] I don't think we actually need bootstrap, and even if we did I don't think it would be the best solution for our application. We've already chosen a tool to streamline styling in our application, Material-UI. Introducing bootstrap could lead to divergent aesthitics for components as we continue to build out app, and Bootstrap isn't necessary for media queries anyway. I was able to remove it from the project and resolve many of the issues listed above, 

Second point is more of a question, but do we have to build sass as a separate task right now? Said another way, will sass auto-compile our work when running yarn start? If not, that makes me concerned; if we're going to start adding CSS we should look into [react-scripts ejecting](https://github.com/facebookincubator/create-react-app#converting-to-a-custom-setup) our project entirely and adding it to our build script.

gregquinn2001:
Great comments Dylan. Let's keep the discussion going on this one. Noranda, what do you think?

noranda:
Thanks for the comments! See below.

[GQ1]: Fixed. This was due to bootstrap overwriting the `navbar` class.
[GQ2]: Fixed
[GQ3]: Agreed, fixed
[GQ4]: I'm not sure what you're asking. The partner logos link to the partner sites. Is this not what we want?

[JA1]: Fixed. This was caused by bootstraps global use of `box-sizing: border-box`, which switches everything to a more intuitive box model by not allowing padding and border to affect an elements width. As this is arguably a better model to use, I corrected the spacing of the element.
[JA2]: Fixed. Same reasoning as JA1.
[JA3]: Fixed. Same reasoning as JA1.
[JA4]: Fixed. Same reasoning as JA1.
[JA5]: Fixed. Styles were overwritten by bootstrap.

[DP1]:
Re: Bootstrap -
This is a fair question and I can see valid reasons for both including it and not including it. I've included it with this work because it's an extremely powerful library that continues to be widely-used in web development today. You're right in that I could have implemented media queries without it. I ended up choosing to add it so that we could use the library for future development if we needed it. Usually, I add Bootstrap early on in a project so that these types of class name conflicts don't occur. I should have checked the rest of the app more thoroughly before making this PR, but I've spent some time cleaning everything up. Material-UI is a great library and while it can be used to handle many things Bootstrap can, it's far less customizable. This may not be something we're worried about now, but it may become more so as the app grows. So again, I can see an argument for and against it. At this point though, the app works nicely with both libraries, but if we decide to discard Bootstrap, I can revert it back.

Re: SASS -
This one I'm a little more adamant that this is the approach we should take. The benefits of using a pre-compiler for CSS are huge and most definitely outweigh the inconvenience of having to run two separate tasks. I would advise against ejecting react-scripts, as it can't be undone and it adds a level of complication we don't need. A third option may be to use a package like https://www.npmjs.com/package/concurrently that allows concurrent tasks to run, although I really don't think we need this.

gregquinn2001:
For GQ4 I was talking about the initiatives (shr and ICAREdata) not the partners

gregquinn2001:
For [DP1], should we spend some meeting time to discuss and come to a decision?

noranda:
For [GQ4], there is an SHR link in the description. Is there a link for the ICAREdata study? If so, should I add it to the description as well to keep it consistent? Or keep it as it is but link both logos?

For [DP1], good idea. If we can discuss it in the morning sometime after scrum and we decide to remove Bootstrap, I may have enough time to get it done before the burndown, although I've already been burning somewhat hot on this project this week.

gregquinn2001:
I always try to click on logos but maybe that's just me. Up to you (and Edwin if you want to ask his opinion). The link for ICAREdata is http://icaredata.org/.

nicoleng12:
This looks really good! Very professional. All the tests pass. One small thing:

[NN1] - When I go to full mode I see the following error in the console:
image

noranda:
[NN1]: Fixed!
[GQ4]: Added links to initiatives logo on landing page.
[DP1]: After giving this some more thought, I think having three large libraries (bootstrap, material-ui, and react-flexbox-grid) is probably overkill for our needs. We can always re-evaluate bootstrap in the future if we decide it might be useful, but given I was only using a simple mixin, I decided best to remove it for now. I've created my own media mixin, which can be used for other components once SASS is properly set up.

Thanks!

gregquinn2001:
Wow Noranda! Sounds like it may make sense.

Dtphelan1:
I want to be clear -- I'm not saying we don't use SASS. I like SASS and have used it on other projects. I'm saying the way we are using it right now does not integrate with our current workflows and will add extra steps that can lead to errors in building and deploying. I definitely think the project will value from our using SASS, it just needs to be integrated into the project better.

jafeltra:
This looks great! Thanks for addressing all my JA comments! They all seem to be fixed now. Also, I don't see the warning that Nicole pointed out. 

As for the SASS task, would it be possible to also add the watch-css task into the start script? I think that's how we have it on CDS Connect and I think it works correctly that way and would avoid needing to eject the project. Our start script is just "npm run watch-css && react-scripts start" so both tasks run when you run "yarn start". Would this be a possibility?

noranda:
Whoops, forgot to mention I fixed [NN1]!

@jafeltra : Our CDS Connect start script doesn't actually work (I'm been meaning to mention it). The first command `npm run watch-css` won't exit and thus never call the second command. I've been running them separately. I'm interested to know how you've been managing it!

jafeltra:
Hmm I don't think I knew that. I've only run it separately like twice before. It always tells me it built 1 CSS file and I've had errors in it before and it won't work.

jafeltra:
If we are deciding to figure out a way to run the watch-css task at the same time as the start task at a later point, I think this PR looks good to me! Nice work!

nicoleng12:
I think looking into running the watch-css task can be put as a separate jira (I believe that's what we decided at scrum?). All issues have been addressed. Looks good!

PR #193, opened 2017-11-28
Fix width has been removed; improved horizontal sizing

Dtphelan1:
Small PR so I've removed the majority of the checks we usually include here. Just removes some margins and a max-width cap that as causing weird page formatting on zoomed out screens or large monitors. Give this a test on your ends and let me know if you see anything else we can do in the way of styling.

### DIFF
--- a/src/apps/FullApp.css
+++ b/src/apps/FullApp.css
@@ -5,7 +5,6 @@
 
 /* Restricts app to max fixed width for very wide screens. */
 .FullApp-content {
-    max-width: 1750px;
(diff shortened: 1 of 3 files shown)


gregquinn2001:
It looks like the multi-choice buttons in slim mode might have space between them now? Although it looks like master does too?

danlee1025:
I think torch fixes it in his pull request.

gregquinn2001:
thanks Dan. Julia and Laura corrected me as well.

PR #194, opened 2017-11-28
Update SummaryMetadata to specify type in each section

danlee1025:
- Added type field to each section object in SummaryMetadata.
- Renamed collection field to data in SummaryMetadata.
- Moved 2 column logic from DataSummaryPanel to DataSummaryTable so that each section is divided individually instead of dividing up all sections into 2 columns.
- Added logic in DataSummaryPanel to look at the type of the section to decide what component to render.
     -  Currently only supports type of 'NameValuePairs'.
     -  In the future, we might create a component manager to allow the user to choose different views based on the data type in each section.

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- Cheat sheet is updated
- Demo script is updated 
- Documentation on Wiki has been updated 
- Additional technical components and libraries are documented and added to wiki as needed

**Code Quality**

- Use public class field syntax for binding functions - ex. handleChange = () => { ... };
- [x] Code is commented

**Tests**

- [x] Existing tests passed
- Added UI tests for slim mode 
- Added UI tests for full mode
- Added backend tests for fluxNotes code
- Added note parser examples to test new functionality

## Reviewer 1: Nicole

- [x] The tests appropriately test the new code, including edge cases

## Reviewer 2: Name

### DIFF
--- a/src/summary/DataSummaryPanel.css
+++ b/src/summary/DataSummaryPanel.css
@@ -37,5 +37,9 @@ tr.existing-note-entry td:last-child {
     border-bottom: 1px solid #ccc;
 }
 
+#condition-summary-section h2.section-header{
+    font-weight: bold;
+}
+
 
 
--- a/src/summary/DataSummaryPanel.jsx
+++ b/src/summary/DataSummaryPanel.jsx
@@ -34,12 +34,32 @@ class DataSummaryPanel extends Component {
         return conditionMetadata;
     }
 
+    // renderSection checks the type of data that is being passed and chooses the correct component to render the data
+    // Current implementation only renders the DataSummaryTable
+    // TODO: Render other types of data
+    renderSection(section) {
+        const {patient, condition, onItemClicked, allowItemClick, isWide} = this.props;
+
+        if (section.type === 'NameValuePairs') {
+            return (
+                <DataSummaryTable
+                    patient={patient}
+                    condition={condition}
+                    conditionSection={section}
+                    onItemClicked={onItemClicked}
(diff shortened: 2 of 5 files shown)


nicoleng12:
This looks good Dan. All tests pass. Nice job moving the division of columns into the specific section.

jafeltra:
This looks good to me! All the tests pass and adding a data type of NameValuePairs seems like a good approach to me!

The one question that I have is about splitting the tables into two columns at the individual section level. I talked to Nicole about this so I understand the benefits of doing it this way so that the full section stays together and you have more control over each section individually. But it seems like now we have a lot more unused space on the right side of the screen since the majority of our tables only have one subsection. The designs for this might still be changing so I don't think this is an immediate problem, but I figured I'd mention it.

gregquinn2001:
@jafeltra . good question. I think we should monitor it but my assumption is the amount of data will only increase...

danlee1025:
Just rebased the branch...

@jafeltra Yeah I agree it doesn't look as great with the amount of data we have now.  I mentioned to Greg this morning that it looked "ugly".

We would have faced problems in the future when we added sections that wanted to span the whole screen on its own(e.g. Timeline).  With how it was previously, it would have shrunk it into one column.

jafeltra:
@daniellee1025 and @gregquinn2001 thanks for the answers! I think what you both said makes sense. It definitely seems much more useable and flexible this way, and you're right we will probably get more data. So as long as we keep an eye out for it and are aware if it, I'm good with keeping it this way for now.

PR #209, opened 2017-12-14
Implement data view buttons

jafeltra:

This moves the visualization options for the Targeted Data Subpanel onto each individual section and gives the icons functionality. 

Right now, the default option is specified on SummaryMetadata, although this can be moved. I left it there for now because I thought maybe individual sections would prefer a default of one visualization or another, rather than all sections of a certain type having the same default. This can be changed though if it's not the desired approach. Also, I have it so that even sections with only one visualization option give an icon showing what option is displayed. I thought that it provided a little more context at what visualization was provided, but I'm open to change this if others feel differently.

This also fixes the backend test to check that the state updates when the icons are clicked. I added a UI test to check the right visualization is actually rendered (with lots of help from lots of people! Thanks!).

## Submitter:

**Running the Application**

- [X] Manually tested in IE

**Documentation**

- [] **N/A** Cheat sheet is updated
- [] **N/A** Demo script is updated 
- [] **N/A** Documentation on Wiki has been updated 
- [] **N/A** Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- [] **N/A** Note parser has been updated if structured phrases change

**Code Quality**

- [X] Code is commented

**Tests**

- [X] Existing tests passed
- [] **N/A** Added UI tests for slim mode 
- [X] Added UI tests for full mode
- [X] Added backend tests for fluxNotes code
- [] **N/A** Added note parser examples to test new functionality

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/src/summary/NarrativeNameValuePairsVisualizer.jsx
+++ b/src/summary/NarrativeNameValuePairsVisualizer.jsx
@@ -93,6 +93,9 @@ class NarrativeNameValuePairsVisualizer extends Component {
     */
     buildNarrativeSnippetList(template, subsections) {
         let result = [];
+        if (template === undefined) {
+            return result;
+        }
         const len = template.length;
         let index, start = 0;
         let pos = template.indexOf("${"), endpos;
--- a/src/summary/NarrativeNameValuePairsVisualizer.jsx
+++ b/src/summary/NarrativeNameValuePairsVisualizer.jsx
@@ -160,6 +160,9 @@ class NarrativeNameValuePairsVisualizer extends Component {
         let subsections = this.getSubsections();
 
         let narrativeTemplate = "";
+        if (conditionSection.narrative === undefined) {
+            return [];
+        }
         conditionSection.narrative.forEach((sentenceObject) => {
             const template = this.getTemplate(subsections, sentenceObject);
             narrativeTemplate = narrativeTemplate.concat(template).concat(". ");
--- a/src/summary/SummaryHeader.css
+++ b/src/summary/SummaryHeader.css
@@ -39,5 +39,6 @@ h2, h3 {
 
 .small-btn {
     min-width: 0 !important;
+    min-height: 0 !important;
     margin-left: 1px;
 }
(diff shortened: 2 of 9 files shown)


gregquinn2001:
Now I see what you meant by default visualizer @jafeltra . I would prefer to just default to the 1st one. The sections should not know anything about what visualizers exist. Visualizers are an extension point and decoupled from the defined summaries.

danlee1025:
[DL1]  If I switch to the narrative view for the summary and then change the condition to Fracture, I get a console error.  I'm guessing this is because the Fracture condition doesn't have a narrative view so we might need some null checking there.
[DL2] This might be related to DL1.  If I select the narrative view for Pathology Results and then change the condition to Fracture, the timeline does not appear by default.

jafeltra:
I removed the default visualization out of summary metadata and instead made the default the first of the visualization options that are available. I believe this addresses Greg's comment. I'll look at Dan's comments now.

jafeltra:
Just pushed a commit that added a check for undefined in the Narrative visualizer component and added a second case for resetting the default visualizer in the TargetedDataSection component. This should fix DL1 and DL2. Let me know if anyone finds anything else!

gregquinn2001:
This is so cool to see this working. Nice job!

[GQ2] Sorry to be difficult, but really the TargetedDataPanel shouldn't know anything about tabular vs. graphical vs. narrative vs. visualization xyz. It should only know about visualizers in general. Looking at the code, I'm seeing a lot of references to tabular and narrative right now. Is that just naming or does that code still know about narrative vs. tabular vs. graphical (e.g., timeline)? For example, it looks like the SVG icons are controlled within TargetedDataSection. I think visualizers should provide their icons. We can go with what we have for now and enter a JIRA for this though since functionality it's awesome!

gregquinn2001:
[GQ3] Again, we can merge this as is and create a JIRA if we think this warrants a possible change. Should an icon be displayed next to a section header if there is only 1 visualizer?

PR #222, opened 2017-12-28
New allergies inserter

gregquinn2001:
JIRA 817.

Added some allergies to hard coded patient (on client side only right now). Implemented @allergies shortcut that inserts a summary of allergies for the patient. Just outputs names right now. Unfortunately, it looks like latest SHR model has changed substantially on web site vs. what our generated classes are.  As soon as ES6 classes have deserialization methods from SHR JSON and from FHIR, we can update which could be a lot of work.

## Submitter:

**Running the Application**

- [X] Manually tested in IE

**Documentation**

- [N/A ] Cheat sheet is updated
- [N/A ] Demo script is updated 
- [N/A ] Documentation on Wiki has been updated 
- [N/A ] Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- [N/A ] Note parser has been updated if structured phrases change

**Code Quality**

- [X] Code is commented

**Tests**

- [X] Existing tests passed
- [N/A ] Added UI tests for slim mode 
- [N/A ] Added UI tests for full mode
- [N/A ] Added backend tests for fluxNotes code
- [N/A ] Added note parser examples to test new functionality

## Reviewer 1: Nicole

- [x] The tests appropriately test the new code, including edge cases

## Reviewer 2: Name

### DIFF
--- a/src/dataaccess/HardCodedPatient.json
+++ b/src/dataaccess/HardCodedPatient.json
@@ -466,5 +466,14 @@
 			"whenClinicallyRecognized": {"generalizedTemporalContext": { "timePeriodStart": "13 JUN 2017" } },
 			"originalCreationDate": "13 JUN 2017",
 			"lastUpdateDate": "13 JUN 2017"
-		}
+		},
+        {
+			"shrId": "788dcbc3-ed18-470c-89ef-35ff91854c7d",
+			"entryId": "28",
+			"entryType": 	[ "http://standardhealthrecord.org/allergy/AllergyIntolerance" ],
+			"focalSubject": {"entryType": "http://standardhealthrecord.org/demographics/PersonOfRecord", "shrId": "788dcbc3-ed18-470c-89ef-35ff91854c7d", "entryId": "1"},
+			"allergenIrritant": {"value": {"coding": [{"value": "412533000", "codeSystem": {"value": "http://snomed.info/sct"}, "displayText": "Wheat Bran"}]}},
+			"originalCreationDate": "13 JUN 2015",
+			"lastUpdateDate": "13 JUN 2015"
+        }
 ]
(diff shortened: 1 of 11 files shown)


nicoleng12:
Nice! All tests pass and it works in both chrome and in IE. Just one small thing:

[NN1] - All of the inserters we have insert right at the cursor. @allergies does too, but each of the following allergies is displayed on a new line. So it looks like this:

image

which looks a little strange.

Should the allergies be listed on the same line, separated by a comma or if we go with a list, should the insertion insert allergies on a new line instead of at the cursor? Just a thought.

gregquinn2001:
I struggle with this one too. I modeled it after sample notes I've seen where it has an Allergies section and then the allergies are listed within that one per line. I wish we could support both somehow... Any ideas on the best way to handle this? A user can't insert a block of text and then make any changes within it (like adding new lines) so that's not an option right now.

nicoleng12:
I'm not sure what the best way to handle it is. At first I was thinking of just inserting a new line before inserting the block of text, but like you mentioned, if the user makes a new line then inserts, there will be an extra blank line.

Should we leave it as is and expect the user to realize that they need to insert a new line before hitting the inserter? I feel that it is something they would learn quickly from using the inserter and seeing the behavior.

gregquinn2001:
And it's something they can go back after the fact and do as well... not ideal but I'm not sure how else to handle it

gregquinn2001:
Added JIRA 829 to support a conditional newline that inserters can use to cause a new line to be inserted if the current insertion point isn't already at the start of a new line. This could be used at the start this inserter's content to put the block of allergies on its own lines and not to start at the end of a current line in progress.

nicoleng12:
Thanks for adding jira 829! Since NN1 will be covered in jira 829, I approve of the changes made in this branch

PR #229, opened 2018-01-09
Banded regions in line chart

nicoleng12:
Banded regions added to line chart for white blood cell lab results

- Updated white blood count cell values in HardCodedPatient.json so that values cross all three bands in the line graph 
- Used referenceArea in recharts lib to create the bands
- line chart adds bands if the bands attribute is provided in summary meta data, otherwise the the chart is drawn without bands

Known issues/things to look into:
- Values for the white blood cell count need to be verified by a clinician for what the values should be for a typical breast cancer patient
- Band range values (low and high) need to be verified by a clinician. Current band ranges used are based on a quick google search for normal white blood cell values 

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- N/A Cheat sheet is updated
- N/A Demo script is updated 

**NoteParser**

- [x] Note parser has been updated if structured phrases change

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed
- N/A Added UI tests for slim mode 
- N/A Added UI tests for full mode
- N/A Added backend tests for fluxNotes code

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/src/dataaccess/HardCodedPatient.json
+++ b/src/dataaccess/HardCodedPatient.json
@@ -125,7 +125,7 @@
                     "http://standardhealthrecord.org/observation/Observation",
                     "http://standardhealthrecord.org/base/Action" ],
                 "clinicallyRelevantTime": "25 NOV 2017",
-                "value": {"value": 1.74, "units": {"value": "10^9/L"}},
+                "value": {"value": 6.74, "units": {"value": "10^9/L"}},
                 "specificType": {"value": {"coding": [{"value": "C0023508", "codeSystem": {"value":"http://ncimeta.nci.nih.gov"}, "displayText": "White blood cell"}]}},
                 "status": "final",
                 "_issue": "strictly speaking, the observations in this list are elements and not entries so entryType shouldn't exist but how would you know which concrete type this represents otherwise? Problem is bigger for choices where some or all choices are just elements"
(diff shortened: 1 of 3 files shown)


nicoleng12:
@danlee1025 pointed out that the way I had set up the code, it was limited to just 3 bands. He and I talked and he proposed some fixes that allows for any number of bands as well as the case where bands might have the same color (in the mockups, cholesterol has 2 "bad" regions that are red that surround a middle good region). I have incorporated Dan's edits which make the bands more versatile for any type of lab data we might have to plot

gregquinn2001:
Thanks @danlee1025 ! @nicoleng12 this looks really good to me!

gregquinn2001:
[GQ1] Why do we need assessment and color as attributes of a band? The assessment should be the only thing provided in the metadata and the visualizer should map that to a color. The metadata should be data that drives visualization and not visualization properties directly if that makes sense.

danlee1025:
For GQ1, Nicole and I discussed it a little bit.  I had added color attribute just for testing but Nicole thought it would be good to leave it in for more versatility in case we wanted different colors in the future.  We can definitely remove the color property if you don't think it belongs in the metadata.

jafeltra:
This looks really good! I have one question that we talked a little about in person, but wanted to also pose to others. Would we want to have a band for the remaining uncolored top portion of the y axis? This might mean that the top band would be another color. It initially seemed odd to me that part of the graph was just white without a band, but I also understand that stopping the green band at 10 and not extending it to the full height of the graph is the best choice since a value above 10 might not be considered "good" anymore. Would we want to have a fourth band that goes from 10 to the max height of the graph? I know Nicole had said this might be difficult with the Recharts library, so this could be a separate Jira, if we want it at all.

[JA1] Can we combine the if else statement that renders the LineChart so that we only have it once and just have the renderedBands return null if the graph doesn't have any bands? I think this would help consolidate some of the code.

nicoleng12:
Great suggestions everyone! For [GQ1] I moved the colors out of summary meta data and put the case statements back into the banded line chart visualizer. @gregquinn2001  I agree that we should keep chart styling withing the chart component. @danlee1025 We can still add additional colors if there are more bands or if some charts need different colors. We can handle this from inside the chart component. 

For [JA1] I followed Julia's suggestions which cleaned up the code nicely. Thanks!

jafeltra:
This looks good to me! After talking, we decided to hold off on filling in the top region of the chart until we find out if it is necessary. If it is, we will add a Jira to the backlog. All the tests pass and JA1 is addressed. Nice job!

jafeltra:
Once this is rebased, I can merge it.

nicoleng12:
Thanks! I just rebased

PR #230, opened 2018-01-09
Progression chart

Dtphelan1:
For JIRA 837 - Implement Progression as a line graph visualization

Creates a new visualization component and a new data type for disease status/progression. 

Known issues: 
- While we use Disease Status as the keyword in our application, anything SHR generated uses "progression" instead. This can lead to naming and reference ambiguities throughout our code. Something to discuss so we can define best practices/when variables should be named one after the other. 
- Right now the viz is disabled unless in pre-encounter mode, but because there are issues with resizing viewing this visualization in a demo setting would be impermissible. To see why, see what happens when you go from post- to pre- and back to post-encounter mode -- the WBC will be thrown all out of wack. If we want this vis present in a demo and if we don't fix the resizing, then we may wan to remove the clinicalEvent specific viewing. 

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- N/A Cheat sheet is updated
- N/A Demo script is updated 

**NoteParser**

**Code Quality**

- [x ] Use public class field syntax for binding functions - ex. handleChange = () => { ... };
- [x] Code is commented

**Tests**

- [x] Existing tests passed
- N/A Added UI tests for slim mode 
- N/A Added UI tests for full mode
- N/A Added backend tests for fluxNotes code

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/src/model/oncology/FluxProgression.js
+++ b/src/model/oncology/FluxProgression.js
@@ -1,6 +1,5 @@
 import AssessmentFocus from '../shr/assessment/AssessmentFocus';
 import CodeableConcept from '../shr/core/CodeableConcept';
-import SpecificType from '../shr/core/SpecificType';
 import Evidence from '../shr/observation/Evidence';
 import Progression from '../shr/oncology/Progression';
 import Reference from '../Reference';
@@ -54,6 +53,14 @@ class FluxProgression extends Progression {
             ev.value = lookup.getEvidenceCodeableConcept(e);   
             return ev;
         });
+    }
+
+    /**
+     *  Getter for code
+     *  This will return the code string from CodeableConcept
+     */
+    get code() {
+        return this._codeableConcept.coding[0].code;
     }
 
     // Flux added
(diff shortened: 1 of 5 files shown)


jafeltra:
This looks good to me! I think eventually we may want to have a more generic component for this type of graph (graphing categorical data as quantitative data and doing the necessary conversions), but I think the visualizer that you have works for what we need now. Since the mapping of data will be specific to each piece of data, I think it makes sense that this happens in summary metadata. All the existing tests pass as well.

gregquinn2001:
[GQ1] The code getter on FluxProgression should really be statusCode or statusAsCode since we call the getter for its display value "status"

gregquinn2001:
[GQ2] I think the visualizer should determine the mapping. That's why we made the data type be disease status specific so visualizers would understand the disease status values. I think that what makes this hard to be generic. What we put in the summary metadata should just be data!

Dtphelan1:
As for [GQ2] I think the reasoning for putting the mapping in the metadata was that if we wanted to make it a generic component we would have to specify mapping when we provide values. (I realize now that the names I gave the lookups aren't very generic... whoops). Given that, do we still want to remove the mapping from the metaData? It wouldn't take much at all, I'd just change the assignments in the progressionVisualizer to be hardcoded instead of pulling from the condition.

Dtphelan1:
[GQ2] resolved

PR #231, opened 2018-01-10
added float back; tested manually for issues

Dtphelan1:
Really simple PR, doesn't touch much. I'd suggest pulling down and checking for layout quirks in a variety of different screen sizes.


nicoleng12:
This looks good! I tested it on chrome and IE, on a large screen, on a small screen, zooming in, zooming out, etc.

PR #241, opened 2018-01-24
Revisit insertion from targeted data panel

Dtphelan1:
Addresses ASCODCP-788:

Implements a new approach for insertions of the narrative visualizer's structured-data elements. The model for insertion can be replicated in other visualizers to replace the plus-button mechanism or add insertion behavior where there currently is none.  Behavior is such that hovering over a blue-underlined element for more than 1/2 second will result in a menu of actions users can take relating to that element. Currently there is only one action -- insert a text represetnation of the structured-data value -- but more menu items can be added in the future.

Known shortcoming, which reviewers should remedy if possible: 
1. The position of the dropdown is fixed to the mouse location at the time of first touching/entering the underlined text. 
2. If you hover over the text, trigger the dropdown, then leave the text without ever hovering over/entering the dropdown, the menu doesn't fade away on a timer; the only way of closing the menu is to click outside of the dropdown. 

Additional changes: 
1. This also fixes the i.e. scroll bar bug noted in ASCODCP-787. The bug was a major hindrance on this new feature's UX in IE, and was taken care of accordingly.
2. This PR updates our material UI version to beta-30. This was necessary to use absolute positioning of the dropdown instead of relative positioning anchored to the div of the underlined text. The latter approach produced ugly behavior on multi-line structured-data snippets. 

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- **N/A** Cheat sheet is updated
- **N/A** Demo script is updated 
- [x] Documentation on Wiki has been updated 
- **N/A** Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- **N/A** Note parser has been updated if structured phrases change

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed
- **N/A** Added UI tests for slim mode 
- **N/A** Added UI tests for full mode
- **N/A** Added backend tests for fluxNotes code
- **N/A** Added note parser examples to test new functionality

## Reviewer 1: Name

## Reviewer 2: Name


jafeltra:
This looks really awesome! I think the click to dismiss the popup is natural so it does not stand out to me. The popup feels like it opens at a natural place as well, whether you're at the end of a line or not and it does not block any of the text that you are currently looking at. I didn't notice any unintended changes with the upgraded Material UI version. Awesome job!

nicoleng12:
Everything looks great! All tests pass. This works on IE. 

[NN1] - Can you add a UI test to test insertion from narrative view?

Dtphelan1:
[NN1] - So.... I don't think Testcafe is smart enough to do what we need it to do. Below you can see how large the divs of narrative snippets are (below is the div for the hemoglobin text snippet). Because of this, test-cafe will technically 'hover' over the element without hovering over the blue-text. This can produce a false red-flag where the dropdown never pops up (because it's not hovering over the text, but it's hovering over the element) even when the dropdown would work otherwise. 

image

nicoleng12:
That's too bad about test cafe. I'm ok then with just merging this since I don't think we want to spend too much time trying to figure out how to get this test to work. What do you think @gregquinn2001 ?

gregquinn2001:
I agree!

PR #249, opened 2018-02-01
Progress note template mid year demo

gregquinn2001:
JIRA 873. Created a new template called progress note in templates list that creates the base note like Joey's sample note.

Outstanding issue: How do you have a note template that needs to mention a condition? If we want to do @hpi then we need a condition. Right now I made the template assume our patient's condition but that is not a long term solution. I will create a jira to address this longer term.

To test: go to new note then click templates in right context context tray and choose progress note. You should see the new note get populated including hpi, active meds, and allergies. Once we have the @reason for next visit, it should populate that as well.

## Submitter:

**Running the Application**

- [ ] Manually tested in Chrome
- [ ] Manually tested in IE

**Documentation**

- [ ] Cheat sheet is updated
- [ ] Demo script is updated 

**NoteParser**

**Code Quality**

- [ ] 4-space indents - convert any tabs to spaces
- [ ] Use public class field syntax for binding functions - ex. handleChange = () => { ... };
- [ ] Code is commented

**Tests**

- [ ] Existing tests passed
- [ ] Added UI tests for slim mode 
- [ ] Added UI tests for full mode
- [ ] Added backend tests for fluxNotes code

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/src/context/ContextTray.jsx
+++ b/src/context/ContextTray.jsx
@@ -15,7 +15,7 @@ export default class ContextTray extends Component {
             value: 1,
             lastActiveContextCount: 0,
             templates: [
-                { name: 'progress note', content: 'REASON FOR VISIT:\n@reason for next visit @condition[[Invasive ductal carcinoma of breast]]\n\nHISTORY OF PRESENT ILLNESS:\n@HPI\n\nREVIEW OF SYSTEMS:\n\nALLERGIES:\n@ALLERGIES\n\nMEDICATIONS:\n@all medications\n\nPHYSICAL EXAM:\n\nASSESSMENT:\n\nPLAN:\n\n' },
+                { name: 'progress note', content: 'REASON FOR VISIT:\n@reason for next visit @condition[[Invasive ductal carcinoma of breast]]\n\nHISTORY OF PRESENT ILLNESS:\n@HPI\n\nREVIEW OF SYSTEMS:\n\nALLERGIES:\n@ALLERGIES\n\nMEDICATIONS:\n@active medications\n\nPHYSICAL EXAM:\n\nASSESSMENT:\n\nPLAN:\n\n' },
                 { name: 'op note', content: 'op note' },
                 { name: 'follow-up', content: 'follow up' },
                 { name: 'consult note', content: '@patient presenting with ' }
--- a/src/context/ContextTray.jsx
+++ b/src/context/ContextTray.jsx
@@ -15,6 +15,7 @@ export default class ContextTray extends Component {
             value: 1,
             lastActiveContextCount: 0,
             templates: [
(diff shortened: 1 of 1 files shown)


jafeltra:
This looks good! The template loads and you can continue adding text into the note.

[JA1] There seem to be a few issues when reloading the template. I'm not sure if this is related to this template or templates in general so I'm not sure if we want to address this as part of this task. If I add a structured phrase into the template, then open a different note, then reopen the note I started with the template, the structured phrase and any plain text after it is deleted from the note. This also happens if I open the template and do nothing in the editor, then open a different note and reopen that note. I can make a Jira if this isn't something we want to address now.

The other thing I noticed isn't related to this task but is more obvious now with a longer note so I figured I'd note it. If you try to add a structured phrase that triggers a popup with a lot of options, the popup opens past the end of the screen. You can scroll down and see the options still but it is a little strange. I can make a Jira for this.

gregquinn2001:
@jafeltra these are great issues to discover now. The 2nd one especially could affect the mid-year review. Can you capture both as jiras? For the first one, it sounds like a general problem with templates and saving notes? I don't think it is specific to this template. If that's not the case, let me know and I'll look into now before proceeding with this pull request.

jafeltra:
For JA1, I can't reproduce it with saving regular notes, but I can reproduce the second case (b) I described in JA1 with the consult note template, but it seems to save structured phrases I add in the middle correctly (a). (I updated my first comment with (a) and (b) parts).

gregquinn2001:
I reproduced JA1 with the other template that has a structured phrase so it's a general problem not specific to the new template. Let's add it as a jira. Unless I'm missing something?

gregquinn2001:
let me know if we should just discuss via voice @jafeltra

jafeltra:
I think you're right that it's general to templates. Even though I can't reproduce (b), I think it's probably best to just look at all of it as a separate Jira.

jafeltra:
One other thing I noticed while trying out all those cases was that after choosing a template that does not set a context, should we switch back to the Patient tab in the context tray? This would also be a separate Jira and not something we'd run into for the mid year demo so it's not high priority.

gregquinn2001:
I think your last comment is an area we have to think more about (call it JA3 - maybe go update your comment with a JA3 identifier?). I do think the basic behavior after selecting a template should be to switch back to the patient context or whatever context is most recent. We may have other behaviors required though where they choose list type inserter values somehow... I think it would be good to create a JIRA for this too.

jafeltra:
I have made 3 Jiras for each item mentioned here:

- JA1: Jira 913 (https://jira.mitre.org/browse/ASCODCP-913)
- JA2: Jira 914 (https://jira.mitre.org/browse/ASCODCP-914)
- JA3: Jira 915 (https://jira.mitre.org/browse/ASCODCP-915)

Since these are separate Jiras, this PR looks good to me!

PR #251, opened 2018-02-01
Reason for visit

mtorchio:
This adds a Reason for Upcoming Visit section at the top of the targeted data panel. This is only visualized in narrative mode and allows for insertion. The content is pulled from new EncounterRequested entries in the hard coded patient. Does the underlining of the entire sentence look good and make it clear that it is clickable?

This was rebased to master as of this morning to take advantage of the new ES6 classes, but will likely need to be rebased again before merging.

## Submitter:

**Running the Application**

- [X] Manually tested in IE

**Documentation**

- [**N/A**] Cheat sheet is updated
- [**N/A**] Demo script is updated 
- [**N/A**] Documentation on Wiki has been updated 
- [**N/A**] Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- [**N/A**] Note parser has been updated if structured phrases change

**Code Quality**

- [X] Code is commented

**Tests**

- [X] Existing tests passed
- [**N/A**] Added UI tests for slim mode 
- [**N/A**] Added UI tests for full mode
- [**N/A**] Added backend tests for fluxNotes code
- [**N/A**] Added note parser examples to test new functionality

## Reviewer 1: Name

## Reviewer 2: Name


gregquinn2001:
[GQ1] Can you create a JIRA for the reason for visit to only check for the next future encounter associated with the current condition
[GQ2] Can you create a JIRA for the reason for visit to be left out or indicate "no upcoming appointment for <condition>" if there is no future appointment for the selected condition

gregquinn2001:
[GQ3] In PatientRecord we have getX methods and getXChronologicalOrder methods. In some cases we also support an optional argument to filter X to only be ones since a specified date/time. Can't we just call getEncountersChronologicalOrder(new moment) then use the first one returned if any? Eventually that will also support accepting a condition to filter on but that's one of the above JIRAs. Examples in PatientRecord right now: getMostRecentProgressionForCondition (for optional since date) and getProgressionsForConditionChronologicalOrder which sorts chronologically.

mtorchio:
I've created JIRAs for GQ1 and GQ2, and I'm updating PatientRecord to add and use a getEncountersChronologicalOrder with no parameters.

gregquinn2001:
For GQ3, that function would just return all encounters in chronological order. Are you going then use that within your getNextEncounter function in patient record?

mtorchio:
Yes. getNextEncounter() can't just return the first item off that list, because it may be in the past.

So either that complexity can live in a getUpcomingEncountersChronologicalOrder, or stay in getNextEncounter.

gregquinn2001:
In your first post you said you were just adding a method called getEncountersChronologicalOrder. That method wouldn't filter at all. If you're adding one called getUpcomingEncounterChronologicalOrder then that is different and would filter by today's date. I'd prefer a more generic method that accepts a date like we've used elsewhere but I guess that's fine.

mtorchio:
Oh, I misunderstood what your mention of "getEncountersChronologicalOrder(new moment)" in GQ3 implied. That would generically accept a date and return a list of encounters on the same day or after that date. If new moment() is passed in, it will use today's date as the cutoff. Then getNextEncounter could just return the first item on that list.

I won't be making a getUpcomingEncounterChronologicalOrder then, it's not needed.

gregquinn2001:
Yes but either way is fine. The date filtering can occur in the getNextEncounter method too. If you do have a generic method that accepts a date and only returns encounters scheduled for after that, it should actually be a date and time. For encounters on the day of, we want to show a 3pm appointment if it is 2:45pm because that's a very likely scenario!

mtorchio:
Pushed a fix for GQ3. Moment is very picky about date formats but HardCodedPatient.json is now using the correct one for date+time.

gregquinn2001:
[GQ4] you can tell moment what format the date/time you are parsing is in. I'd prefer our date/time format is the same as our date format but with a time as well.

new moment(dateString, "D MMM YYYY h:mm a");

gregquinn2001:
[GQ5] I'm creating a jira against the minimap too since the name of the section doesn't fit in the space given the section in the minimap. it should truncate I guess...

gregquinn2001:
[GQ6] Can we shorten the name of the section as a workaround for now? How about "Visit Reason"?

nicoleng12:
This looks good. I also noticed that the minimap issue where "Reason for Upcoming Visit" doesn't fit in the space. GQ6 should fix that. The tests fail, but I think rebasing should fix that

gregquinn2001:
Resolve GQ4 and GQ6. GQ5 was a jira so it's all set.

nicoleng12:
Looks good! All tests pass

PR #258, opened 2018-02-08
Reposition portal at bottom

jafeltra:

This pull request adds a scroll to the suggestion portal and repositions the portal to display upwards instead of down when typing near the bottom of the screen.

While working on this task, I noted some other bugs that were unrelated to this task specifically, but were related to the editor and the template. I haven't addressed these yet and can make Jira items if we think they are issues and are separate.
- Selecting @condition does not actually close the context portal. This is only obvious if you select condition after only typing @c. Once you type @co, the list is small enough that it is hidden behind the next portal.
- Inserting a structured phrase at the very bottom of the progress note template still causes the location of the portal to be off. I think this has to do with the way text is inserted from a template because I can't reproduce this by just typing in the editor.
- Inserting a structured phrase in the middle of a template does not trigger the suggestion portal, unless you type a character that we are not checking for in the suggestion-portal.js file (such as `). This happens in the convertSlateDataObjectToCharacter() method. 

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

-  Cheat sheet is updated
-  Demo script is updated 
-  Documentation on Wiki has been updated 
-  Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

-  Note parser has been updated if structured phrases change

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed
-  Added UI tests for slim mode 
-  Added UI tests for full mode
-  Added backend tests for fluxNotes code
-  Added note parser examples to test new functionality

## Reviewer 1: Name

## Reviewer 2: Name


nicoleng12:
I like the scroll bar! In the future, if there are a lot of options in the drop down, this will prevent the drop down from getting too big which is nice.

Tested this in both IE and chrome and it works. Tests pass. I noticed the bugs that you listed above in the description. Can you go ahead and make jira items for those?

jafeltra:
I'm actually looking at the second and third bullets as part of the fix for ASCODCP-913. If they are not resolved as part of that task, I'll make Jira items for them (I also noted that in the Jira task comments). I'll check to make sure the first bullet isn't already a Jira task and if it isn't, I'll create one.

jafeltra:
I've made a Jira item for the first bullet (https://jira.mitre.org/browse/ASCODCP-933).

The second and third are fixed in my other pull request.

gregquinn2001:
[GQ1] @jafeltra I like that we have a scroll bar now! I did notice though that when I arrow down in the menu, the list doesn't scroll so my selection isn't shown. Can we make it scroll based on arrow keys changing the selection?

jafeltra:
@gregquinn2001 That's a good catch. I'll look into fixing that.

jafeltra:
I just pushed a commit that addressed GQ1. Let me know your thoughts on the scrolling behavior. I think I've looked at it scrolling too much to know if it feels natural or not.

I also added the max-height to the context portal since I hadn't done that before. Now if that portal had enough options in it, it would have the same behavior as the suggestion portal. Also, I rebased the branch.

gregquinn2001:
Scrolling works now, but noticed
[GQ2] Hitting escape doesn't close menu. I'm pretty sure it used to but I just tested on my working branch and escape didn't close autocomplete menu. I'll make a jira. Looks good!

PR #262, opened 2018-02-12
Multi line suggestion bug

Dtphelan1:
Fixed a bug concerning the placement of the suggestion dropdown in the case where a user is typing into a multi-line paragraph. In the process I performed a pretty massive overhaul on the suggestion-plugin code, bringing the styling inline with our current code-quality standards, adding comments where I could, and updating the code in a few other places. As a result, a few other bugs have been addressed in the process. 

1. Original Bug: Shortcuts that require two selections (e.g. '@condition' prompts users for a second selection containing all the patients current conditions) would not close the first suggestion menu. This menu now closes. 
2. Original Bug: Backspacing into a suggestion (e.g. '@age' -> '@ag') used to remove the menu/disable the insertion of a suggestion. This is no longer the case, and the expected element is now inserted.
3. Original Bug: The nil case for suggestions (i.e. just '@') used to always insert the previously typed shortcut, regardless of what the suggestion menu indicated. This has now been fixed.

Hopefully this, along with the work Julia has been doing, should prepare our for the demo and make it easier to make changes to the editor in the long term. 

It's also worth noting that the changes we've made to the suggestion plugin have become pretty drastic. A result of this is that I suggest we use my fork of the original repo (https://github.com/Dtphelan1/slate-suggestions/) as the canonical version of the tool. The previous repo is largely a ghosttown, with PR's in limbo for months now (one of them being ours from August!)

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- **N/A** Cheat sheet is updated
- **N/A** Demo script is updated 
- **N/A** Documentation on Wiki has been updated 
- **N/A** Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- **N/A**  Note parser has been updated if structured phrases change

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed
- **N/A**  Added UI tests for slim mode 
- **N/A**  Added UI tests for full mode
- **N/A**  Added backend tests for fluxNotes code
- **N/A**  Added note parser examples to test new functionality

## Reviewer 1: Julia

-  The tests appropriately test the new code, including edge cases

## Reviewer 2: Name

### DIFF
--- a/src/context/ContextPortal.css
+++ b/src/context/ContextPortal.css
@@ -8,6 +8,9 @@
     background-color: rgb(255, 255, 255);
     margin-top: 30px;
     z-index: 1111;
+}
+
+.scrollable {
     max-height: 212px; /* 6 list elements at 32px each, plus 10px of margin on top and bottom */
     overflow-y: auto;
 }
--- a/src/context/ContextPortal.jsx
+++ b/src/context/ContextPortal.jsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import Portal from 'react-portal'
 import Calendar from 'rc-calendar';
 import ContextItem from './ContextItem'
-import position from '../lib/slate-suggestions-dist/caret-position';
 import Lang from 'lodash'
 import './ContextPortal.css';
 import 'rc-calendar/assets/index.css';
@@ -121,7 +120,7 @@ class ContextPortal extends React.Component {
     adjustPosition = () => {
         const { menu } = this.state
         if (!menu || !menu.style) return;
-        const rect = position();
+        const rect = this.props.getPosition();
 
         if (!rect) {
             // TODO: No positioning to use. Removing style may not be correct.
--- a/src/context/ContextPortal.jsx
+++ b/src/context/ContextPortal.jsx
@@ -201,10 +201,12 @@ class ContextPortal extends React.Component {
         const TYPE_CALENDAR = 1;
         const { contexts } = this.props;
         let type;
+        let className = "context-portal";
         if (Lang.isNull(contexts)) return null;
         
         if (Lang.isArray(contexts)) {
             type = TYPE_LIST;
+            className += " scrollable";
         } else if (contexts === "date-id") {
             type = TYPE_CALENDAR;
         } else {
@@ -219,7 +221,7 @@ class ContextPortal extends React.Component {
                 onOpen={this.onOpen} 
                 onClose={this.onClose}
             >
-                <div className="context-portal" ref="contextPortal">
+                <div className={className} ref="contextPortal">
                     {type === TYPE_CALENDAR ? this.renderCalendar() : this.renderListOptions()}
                 </div>
             </Portal>
--- a/src/context/ContextPortal.jsx
+++ b/src/context/ContextPortal.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import Portal from 'react-portal'
 import Calendar from 'rc-calendar';
 import ContextItem from './ContextItem'
+import position from '../lib/slate-suggestions-dist/caret-position';
 import Lang from 'lodash'
 import './ContextPortal.css';
 import 'rc-calendar/assets/index.css';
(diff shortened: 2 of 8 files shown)


jafeltra:
This looks really awesome! All these changes really help the editor work more intuitively. The original bug, along with the three you listed, all are fixed. I also texted out typing a shortcut that would span two lines (if you type @condition 4 characters from the end of the line), and the portal follows the cursor and gets repositioned in the correct place. All the tests pass.

jafeltra:
[JA1] If you open the condition selection portal, the page seems like it shifts. Dylan and I already talking about this one, but figured I'd note it here. He's looking into it.

Dtphelan1:
[DP1] When users click on muli-value suggestion (e.g. @medication) the portal for options either won't pop up, will pop up above the cursor instead of above where the editor-caret is, or will pop up with the original dropdown containing @medication

Dtphelan1:
JA1 is techincally fixed, but DP1 is a subset of this bigger issue. I'm not going to be able to fix it before I leave, as it has to do with weird interactions between concept of a Context Portal and the suggestion portal. I'd ask the people if they think it'd be worth merging (assuming there aren't other issues) and fixing the existing bugs, while recognizing that there's a new bug that we're introducing.

gregquinn2001:
I'll try to take a look at it, but what do you think @jafeltra? Is the app functionally better for demos despite the new bug? I'm assuming you're saying it is better for demos @Dtphelan1 ?

Dtphelan1:
AFter talking with Julia, I don't think merging this without addressing the new bugs makes a lot of sense. While there are fewer bugs, they are certainly more jarring to the user than the previous ones.

gregquinn2001:
Ok. thanks! Julia, is this something that you can look at after Dylan has left for vacation?

jafeltra:
Yeah I think so. Since the other bugs this fixes are really useful, I think it would be good to try to figure out the remaining bug and I will need a new task soon anyway.

jafeltra:
I've pushed updates that I think fix the remaining bugs. The approach I took to fixing the bugs required passing in new props to the suggestions portal. I thought since we have such significant changes to the portal, it would be okay. If we don't think this is a good thing to do, let me know and we can maybe revisit how the changes are made.

Right now, the portal aligns with the cursor's position in the editor. The second portal, the context portal with the conditions list, gets positioned in the same location when it opens based on using the enter key to select @condition, clicking condition, and autocompleting the word condition. The first portal successfully close when this is opened.

I was also able to close the portal with the escape key as part of this task. The Jira's that I think have been fixed by this PR are below. I don't think this list covers all the bugs Dylan listed at the top of this PR, which should still all be fixed.
- Jira 933: the first suggestion portal closes if you select something that opens the second (https://jira.mitre.org/browse/ASCODCP-933).
- Jira 934: Choosing the initial option after typing @ or # works as expected (https://jira.mitre.org/browse/ASCODCP-934)
- Jira 941: Escape key closes the portal (https://jira.mitre.org/browse/ASCODCP-941)

Dtphelan1:
It looks like @jafeltra and @gregquinn2001 have made some great fixes to the bugs before. I'm not finding any new bugs, but I might be a little blind to the testing in virtue of having done it too many times... Any chance we can get a reaffirmation from @gregquinn2001 and @jafeltra that this task is looking good, and at least one new review (maybe from @danlee1025 or @laclark22?) Others should let me know what you think!

Dtphelan1:
Talking with @jafeltra and she's pointed out another bug that exists, though it's not blocking: 
[JA(-DP?)2] A portal started on one line doesn't overflow off that line in sync with the text. That is to say, if a shortcut is started on line 1 and overflows over onto line 2, the portal lags behind the shortcut in an off-by-one fashion. Eventually it catches up, but no on that first overflowing character.

nicoleng12:
[NN1] - When I click on @condition from the context tray, the drop down pops up in the top left corner of the screen. This happens in both IE and chrome

gregquinn2001:
I'd log a JIRA for [JA(-DP?)2]. You definitely win for most complicated bug identifier in a pull request! That's not a challenge though to create more complicated ones!

Dtphelan1:
I'll save further complications for future PR's 😄

Dtphelan1:
@nicoleng12 Does this address your bug sufficiently?

nicoleng12:
@Dtphelan1 Yes that addresses NN1, thanks! This PR looks good. All the bugs are fixes. All the tests pass. I've tested this manually in IE and chrome as well.

This just needs to be rebased

nicoleng12:
Rebased the branch

PR #268, opened 2018-02-19
Mid year demo patient

gregquinn2001:
JIRA 872. Created a mid-year demo patient in SHR JSON format. HardCodedReadOnlyDataSource can load this patient as well as the default hard coded patient. Updated FullApp to accept a patient id for an endpoint and defined a new endpoint at myd18 that loads the mid-year demo patient.

The mid-year demo patient matches the data that Joey defined within the sample note he created for a metastatic breast cancer patient. All structured data that SHR is capable of representing (or most at least) exists in structured form within the mid-year demo patient json. Recent clinical notes representing the most recent encounters matching the history of present illness in the sample note are also provided in the hard code patient SHR json.

To test, go to http://localhost:3000/myd18.

Testing the regular patient endpoint would be good too.

Added support to display genetics info in targeted data panel.

Also made a number of displays (like medications graphical viz) more robust and handle missing data better.

Also created a number of JIRA issues discovered as part of this.

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- [ ] Cheat sheet is updated
- [ ] Demo script is updated 

**NoteParser**

**Code Quality**

- [ ] Use public class field syntax for binding functions - ex. handleChange = () => { ... };
- [ ] Code is commented

**Tests**

- [x] Existing tests passed
- [ ] Added UI tests for slim mode 
- [ ] Added UI tests for full mode
- [ ] Added backend tests for fluxNotes code

## Reviewer 1: Name

## Reviewer 2: Name


gregquinn2001:
From demo dry run today.
[GQ1] Add dosages for all chemotherapy drugs including at least one out of typical range
[GQ2] Add additional disease status values from past
[GQ3] Pull lab results from Hernandez into Ortiz
[GQ4] Pull Fracture condition from Hernandez into Ortiz

gregquinn2001:
GQ1-5 all addressed.

JIRA 956 also addressed now. Do @condition choosing carcinoma then do @hpi and notice that procedures also include results now.

gregquinn2001:
Added fix for JIRA 946 as well (disease status graph y axis labelling)

danlee1025:
[DL1] On /patient in pre-encounter mode, the disease status graph y values do not all appear.  I think the yTicks need to be sorted(Seems to be the change from master.).  
[DL2] Not a bug on this branch, but in the codeToValueMap in ProgressionLineChartVisualizer, there is a duplicate mapping.  Can you change mapping of "C0677874" to 3.

gregquinn2001:
DL1 and DL2 should be fixed! Thanks @danlee1025

gregquinn2001:
Renamed some methods for consistency within model per @danlee1025 suggestion

gregquinn2001:
rebased

danlee1025:
Added error handling when no review of systems found on patient.  Added review of system entries to myd patient.

jafeltra:
This looks good! I did find a few bugs.

[JA1] @reason for  previous visit is broken in both /patient and /myd18. Inserting the Reason for Visit from the Targeted Data Panel breaks as well.
[JA2] Inserting stage from the Targeted Data Panel breaks on /myd18.

danlee1025:
JA1 should be fixed.  Just had to add the new function name to shortcuts.json.

jafeltra:
JA2 is due to the fact that I was trying to insert the existing staging value before using a #staging creator. The StageInserter has StagingCreator as it's parent context. I talked to Dan and we think the StageInserter should have a ConditionInserter parent context, and then return the stage for the specific condition, if it has a stage. @gregquinn2001 does that seem like the right way to approach fixing this?

jafeltra:
I realized that doesn't actually solve the problem. When you try to insert the staging value from the Targeted Data Panel, you still would need a condition specified before it can be inserted.

gregquinn2001:
Yes! Good call. I agree!

gregquinn2001:
You should need a condition (a breast cancer one specifically)

gregquinn2001:
but a patient can have multiple staging values as well so hence the staging context requirement. @current staging would make sense with condition context only and it would find the lastest staging and get the stage from that

gregquinn2001:
we don't have an @staging so you can't use @stage when you should be able to right now

jafeltra:
That does make sense to me. I'm not sure what the best way to move forward with this is. Should we implement the @current staging shortcut like you said in order to fix this? One quicker fix we could do is take the shortcut off of the Stage section in summary metadata, but that doesn't seem like the best thing to do since it is an inserter and we would want to capture that as structured data. I was also thinking maybe we could add it as plain text if the parent context was not set, but after talking to Dan, I'm not sure that is easy to do.

jafeltra:
Updated the list that gets created for inserters so null values don't get printed out. This was happening when using @all medications/@active medications since the units on timingOfDoses was null. Everything else looks good to me!

PR #279, opened 2018-02-28
Note disappearing when changing encounter modes

mtorchio:
Switching the Workflow dropdown to Post-encounter or Pre-encounter used to cause the editor contents to disappear. This PR resolves that so the editor does not lose contents. Communication between components is accomplished by adding a couple more props at the FullApp level - using global state seems less than ideal to me, there may be a better way to accomplish this.

Running tests now.

## Submitter:

**Running the Application**

- [X] Manually tested in IE

**Documentation**

- [**N/A**] Cheat sheet is updated
- [**N/A**] Demo script is updated 
- [**N/A**] Documentation on Wiki has been updated 
- [**N/A**] Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- [**N/A**] Note parser has been updated if structured phrases change

**Code Quality**

- [X] Code is commented

**Tests**

- [X] Existing tests passed
- [**N/A**] Added UI tests for slim mode 
- [**N/A**] Added UI tests for full mode
- [**N/A**] Added backend tests for fluxNotes code
- [**N/A**] Added note parser examples to test new functionality

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/src/containers/FullApp.jsx
+++ b/src/containers/FullApp.jsx
@@ -67,6 +67,7 @@ export default class FullApp extends Component {
             summaryItemToInsert: '',
             summaryMetadata: this.summaryMetadata.getMetadata(),
             noteClosed: false,
+            currentlyEditingEntryId: -1,
 
         };
     }
--- a/src/dashboard/ClinicianDashboard.jsx
+++ b/src/dashboard/ClinicianDashboard.jsx
@@ -11,9 +11,9 @@ export default class ClinicianDashboard extends Component {
 
         this.state = {
             targetedDataPanelSize: "default",
-            notesPanelSize: "default"
+            notesPanelSize: "default",
+            currentClinicalEvent: "pre-encounter"
         };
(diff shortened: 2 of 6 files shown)


gregquinn2001:
[GQ1] In pre-encounter mode, I hit new note then type in some content. Switch to post-encounter and the note stays. If I close that note and switch back to pre-encounter mode, the in-progress note is re-opened. I don't think it should be
[GQ] With an in-progress note open, switching from post-encounter to pre-encounter maybe shouldn't keep that open but instead close it (it's still in an in-progress note so not lost)?

I think the correct behavior needs to be thought through a little more here.

gregquinn2001:
Closed because we need to think through all the scenarios here (see GQ). When should the open note close? When should a new note be created? Should an existing in-progress note ever be auto-opened?

Should we just eliminate the workflow drop-down and not expose that to the user?

PR #301, opened 2018-03-23
Generic viz actions

laclark22:
Visualizers should support generic actions to be done on structured data - signed, unsigned, and/or missing. Should be able to specify if they can be done only when an editable note is open or not as well. There is an action object in full that is sent to tabular visualizer for inserting data from the summary panel. The idea here is there will be more actions as we continue forward (open source clinical note, options for missing data) so we wanted to pull that out from tab viz and make action objects in full app that are sent to the tab viz and rendered as a menu item depending on type, criteria etc. 

## Submitter: Laura Clark & Dan Lee

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- NA Cheat sheet is updated
- NA Demo script is updated 
- NA Documentation on Wiki has been updated 
- NA Additional technical components and libraries are documented and added to wiki as needed

**NoteParser**

- NA Note parser has been updated if structured phrases change

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed
- NA Added UI tests for slim mode 
- NA Added UI tests for full mode
- NA Added backend tests for fluxNotes code
- NA Added note parser examples to test new functionality

## Reviewer 1: Name

## Reviewer 2: Name


jafeltra:
This looks good! There's a TODO in FullApp.jsx that I think is done that you can remove.

Only thing I found!
[JA1] Inserting text in the narrative does not say the text to be inserted.

gregquinn2001:
Very cool. It will be great to have the ability to add new actions by just coding the actual action it performs and adding metadata! I did a quick look at the code and saw this but I'll check it out more:
[GQ1] the text property should have ${elementText} in the string within the metadata to indicate that the menu item's name should include the text of the element being clicked on. We could support pulling other properties if needed later. Note that most action menu items will probably not have the text of the element in it (e.g., Show Source Note will not). 

[GQ2] I think we can clarify type and signed/unsigned and editable note criteria perhaps. They are all criteria to determine if the menu item applies. I was originally thinking the criteria should be able to say a menu item applies to missing, unsigned, or signed elements, but I think that would be rare so specifying a single type may make sense: value exists or missing and if the value exists the value can be signed or unsigned (and a menu item should be able to apply when the value exists and is signed, unsigned, or either. Finally, the menu item/action could apply only when there is an editable note showing or only if there isn't I guess. So to support that, I think we need something like:

whenToDisplay: {
    valueExists: true or false,
    existingValueSigned: true false or either,
    editableNoteOpen: true or false
}

or something similar that is clearer but still supports all the valid combinations we want...

danlee1025:
I just rebased the branch and will address the comments on the PR.

danlee1025:
JA1, G1 should both be addressed.

I also addressed GQ2 but have a question for @gregquinn2001.  I removed the type property from the actions' object(for now?) as the whenToDisplay property seems to be sufficient to determine when to display specific actions.  What do you think?

gregquinn2001:
Yes. that structure was meant to replace a few of the current elements. I'll check it out. Thanks!

jafeltra:
This looks good! All the comments have been addressed. Just waiting for the tests to pass now.

jafeltra:
Also, just noticing I don't think we don't have Travis set up on this repo.

PR #374, opened 2018-07-12
IE scrollbars

nicoleng12:
This PR addresses jira 787. In IE, there is an extra scrollbar on full app when the editor is open. This is what it looks like on fluxnotes.org currently:

image

This PR removes the extra scrollbar by reducing the height of the app so that there is no scrollbar. The max height is set to 96vh (which is as close as it can get to 100vh before the scrollbar appears). Because the scrollbar issue is not an issue in chrome, the styling change is only applied to IE and Edge so there should be no difference in how the app is displayed in chrome. An important point to note is that I am targeting zoom at 100% in the browser. 

If others have better solutions for how to address the scroll bar issue (instead of introducing a small margin on the bottom of the page), I am open to other approaches. 

Below are approaches I have tried:
- Originally the styling used min-height. I had tried playing around with different values of min-height but in the end only changing min-height to max-height worked
- I looked at this: https://www.sitepoint.com/community/t/extra-scrollbar-when-body-height-is-100vh/257883/2. In index.css, it is already initializing the margin and padding to 0 so this fix did not work
- I had looked into this: https://github.com/vaadin/framework/issues/3364 but the fix to reduce spacing by setting font size to 0 did not work and messed up the styling for the rest of the app
- I looked at this: https://www.bennadel.com/blog/3391-margin-collapsing-causes-unexpected-scrollbar-with-100vh-body-in-webkit.htm and tried to remove any extra padding on components in the app but it did not look like the other components have extra padding. I tried removing the sign button component at the bottom of the app but it had no effect on the extra spacing that was appearing in the app when opened in IE
- I tried using calc(100vh - 40px) to see if the gap is smaller than setting the max-height to 96vh but there did not appear to be a difference in the size of the gap
- I also looked at https://code.tutsplus.com/tutorials/9-most-common-ie-bugs-and-how-to-fix-them--net-7764, specifically the section on "Double Margin on Floated Elements" but the fix did not work and the problem was not quite the same as the one in the app

## Submitter:

**Running the Application**

- [x] Manually tested in IE 

**Documentation**

- N/A Cheat sheet is updated
- N/A Demo script is updated 

**NoteParser**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/src/dashboard/ClinicianDashboard.css
+++ b/src/dashboard/ClinicianDashboard.css
@@ -15,14 +15,14 @@
 /* For IE only */
 @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
     #clinician-dashboard-content .fitted-panel {
-        max-height: calc(96vh - 106px - 16px); 
+        max-height: calc(92vh - 106px - 16px);
     }
 }
 
 /* For IE Edge 12+ only */
 @supports (-ms-ime-align:auto) {
     #clinician-dashboard-content .fitted-panel {
-        max-height: calc(96vh - 106px - 16px); 
+        max-height: calc(92vh - 106px - 16px); 
     }
 }
 
--- a/src/dashboard/ClinicianDashboard.css
+++ b/src/dashboard/ClinicianDashboard.css
@@ -7,8 +7,23 @@
     overflow-x: hidden;
     text-align: left;
     /* 106px is the height of the Patient Control Panel, 16px is the height of the top margin of post encounter view content*/
-    max-height: calc(96vh - 106px - 16px);
+    max-height: calc(100vh - 106px - 16px); 
     min-height: 400px !important;
+
+}
+
+/* For IE only */
+@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
+    #clinician-dashboard-content .fitted-panel {
+        max-height: calc(96vh - 106px - 16px); 
+    }
+}
+
+/* For IE Edge 12+ only */
+@supports (-ms-ime-align:auto) {
+    #clinician-dashboard-content .fitted-panel {
+        max-height: calc(96vh - 106px - 16px); 
+    }
 }
 
 #clinician-dashboard-content .right-border-box {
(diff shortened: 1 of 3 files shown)


jafeltra:
[JA1]: Nicole and I discovered that the scroll bar problem doesn't seem to be fully fixed. I had IE open wide enough so that the top header of the page was the correct height, but the window was not tall enough to view all the shortcuts in the context tray. This caused a scroll bar on context tray, which was correct, but also seemed to create the main scroll bar on the page, which should not be needed. I really don't have a good idea about why those two things are related, but I can't get the main scroll bar to come up in a case where the context tray does not have a scroll bar. Adding a shortcut that increases the list in the context tray (for example, #enrollment) also creates a context tray scroll bar (correctly), but adds the main page scroll bar (incorrectly).

danlee1025:
For JA1, I changed the max-heights from 96vh to 92vh.  I posted screen shots below of how they look on IE.

96vh:
screen shot 2018-07-17 at 2 47 00 pm



92vh:
screen shot 2018-07-17 at 2 50 01 pm


The extra fullapp scrollbar no longer shows up but the margin at the bottom has been increased.  What do others think of this change to fix the scrollbar issue?

jafeltra:
This seems good to me! I haven't noticed the scroll bar come up with the 92vh value. I think the extra white space on the bottom is okay, especially since it only happens for IE.

I did notice that if we only change the NoteAssistant css to use the 92vh, the scroll bar doesn't seem to show up either. I'm not sure if it's better to be consistent with the margin at the bottom, though.

@Dtphelan1 do you have any thoughts about which way is better? If not, I'm good with keeping it as is and I'll approve!

danlee1025:
Pushed up a new branch [ie-scrollbars-2](https://github.com/FluxNotes/flux/tree/ie-scrollbars-2)  for comparison.  The changes on `ie-scrollbars-2` hide the scrollbar until a specific height or width is reached.

jafeltra:
I think the ie-scrollbar-2 branch looks good! The scrollbar only came up when I needed it to. I think since there isn't any white space added at the bottom of the app, we might want to merge that one instead.

danlee1025:
Closing this PR since [another PR](https://github.com/FluxNotes/flux/pull/376) has been made to fix the issue.

PR #406, opened 2018-08-17
Update note header layout

nicoleng12:
This PR addresses jira 1275. I updated the note header to align with more recent concepts (v14.11, page 2). The new design leaves more room for the note name so I do not think we need jira 1238, truncating long note names so POC doesn't overlap (tablet). But I would like to see what others think as well before we delete jira 1238.

In this jira I also created a sass file for FluxNotesEditor styling. When converting the css to scss, I noticed that there was some styling that is no longer being used so I removed those lines after checking that no where else in the app are the ids and classnames being used.  I noticed that the FluxNotesEditor.css contained styling for the suggestion-portal. I left this in the Flux Notes styling sheet because removing it caused styling issues. Ideally the styling for the suggestion portal should be in a styling sheet for suggestion portal since Flux Notes Editor does not use any of the elements that is specified. I suspect that there may be other cases in the app where elements are using styling defined in another class and other cases where styling contains old code so I made a jira (https://jira.mitre.org/browse/ASCODCP-1297) 

## Submitter:

**Running the Application**

- [ ] Manually tested in Chrome
- [ ] Manually tested in IE

**Documentation**

- N/A Cheat sheet is updated
- N/A Demo script is updated 

**NoteParser**

**Code Quality**

- N/A Use public class field syntax for binding functions - ex. handleChange = () => { ... };
- [x{ Code is commented

**Tests**

- [x] Existing tests passed

## Reviewer 1: Name

## Reviewer 2: Name


nicoleng12:
Nice catch @daniellee1025 ! I addressed DL1 and DL2. I'm not sure why git thinks I changed the whole file. I only changed the column widths and the ordering of the note information. It might be because i used the format command in vs code to make sure everything was indented properly

danlee1025:
[DL3]  Talked with Nicole and she will restore some classes that she removed.  Make sure classes deleted from `FluxNotesEditor.css` are actually not being used(check for single quotes).

nicoleng12:
DL3 has been fixed.

mgramigna:
This looks great!

[MG1] Very similar to DL1. When the note name is being edited, the text is a little too close to the scrollbar/left panel. 

image

The TextField for the note name input has the id "note-title-input" for styling references.

src/notes/FluxNotesEditor.scss
```diff
@@ -0,0 +1,161 @@
...
+    -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+    box-shadow: rgba(0, 0, 0, 0.12) 0px 1px 6px, rgba(0, 0, 0, 0.12) 0px 1px 4px;
+    transition: all 450ms cubic-bezier(0.23, 1, 0.32, 1) 0ms;
+    background-color: rgb(255, 255, 255);
+    z-index: 1111;
+    max-height: 212px; /* 6 list elements at 32px each, plus 10px of margin on top and bottom */
+    overflow-y: auto;
+}
+
+.suggestion-portal ul {
```
jafeltra: [JA1] Should we nest these styles like the #clinical-notes styles above it? Or is that part of the separate Jira you created?

jafeltra:
I'm not sure why, but when I regenerate the CSS file from the SASS file, I get minor differences, so git is tracking that the file has changed. I'm not sure why this is happening though. Has anyone else seen this?

mgramigna:
@jafeltra I think this might have to do with how line endings are treated on Mac vs. Windows. I'm not 100% positive. But when I was doing SASS the behavior didn't happen, but I think others have mentioned that it happened with Dylan's SASS. It's probably ok to ignore for now since when we fully integrate SASS, all .css files will be in the .gitignore anyway.

jafeltra:
@mgramigna That's what I was thinking also, so I'm okay with ignoring it for now. I did also get this difference, which I don't think matters much either, but doesn't seem to be related to line endings: `-webkit-tap-highlight-color: transparent;` changed to `-webkit-tap-highlight-color: rgba(0, 0, 0, 0);`.

nicoleng12:
Thanks guys! Addressed all the issues above (MG1, MG2, JA1). As for why the css file keeps changing, I'm not sure why. Dan was noticing this too. I added the css files that have been converted to scss to git ignore

jafeltra:
This looks good to me! The header looks good even in portrait mode on the iPad in chrome. Adding the point of care panel doesn't seem to overlap anymore, so I think I would be okay with removing Jira 1238. All the other issues are addressed. Nice job!

mgramigna:
I agree with @jafeltra that giving this much space to notes names inherently solves Jira 1238 for reasonable length note names.

PR #445, opened 2018-09-11
Search structured data

mgramigna:

Jira 1253: Update search to include results in structured data. Worked with @gregquinn2001 and @karenzheung.

Expanded searching to look for basically anything on the left side:

- Can search for values, e.g. "Stable" in the Summary section
- Can search for descriptor, e.g. "Disease Status" in the summary section
- Can search for section names, e.g. "Summary"
- Can search for subsection names, e.g. "Key Toxicities"
- Can search for dates or values on the graphs/timeline
- Searching for the content of a clinical note remains unchanged

Results in the search bar are presented with Section > Subsection in the header and Key: Value in the content. The matching text of the result with what was typed will be bold.

Scrolling:

- Hovering over a selection will jump to the section that the data is contained in
- If the subsection that it is contained in is not visible, it will jump to the subsection
- Note: this also seemed to resolve Jira 1301

We changed a lot for this pull request, so when testing please try to test functionality with the targeted data panel that may not be directly related to search (viewing source, inserting into an open note, etc.).

## Submitter:

**Running the Application**

- [ ] Manually tested in IE

**Documentation**

- [ ] Cheat sheet is updated
- [ ] Demo script is updated 

**NoteParser**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewer 1: Name

## Reviewer 2: Name


jafeltra:
This looks really good to me! I've been testing interactions with the targeted data panel and everything looks really good. One thing I've noticed is that I get a lot of additional errors in the console since I don't have the Treatment Options server running. Is that okay to have?

gregquinn2001:
Thanks Julia! Could you paste a log of your console showing what you're getting? There are some but we definitely don't want a lot if we can help it. Do you get the same on fluxnotes.org though or did our changes for search cause more log messages for treatment options?

mgramigna:
@gregquinn2001 Since the data tries to be re-indexed when the treatment options section re-renders, it will make an API request to re-gather the data. The console errors that show up are the same you get when the server is not running, just repeated each time the treatment options section re-renders

jafeltra:
Matt beat me to it. The error I'm getting is the same failed GET request that we get on the initial page load if the server isn't running.

gregquinn2001:
ok I was assuming that as well but wanted to see if that was the case. Let's put a JIRA in to see if we can't suppress some of the messages but it shouldn't affect someone using the app with the console hidden. Thanks @mgramigna and @jafeltra !

jafeltra:
Sounds good! That's the only thing I've noticed! Everything else looks really good!

nicoleng12:
This looks really good! I like that as you hover over the search results, the app jumps to the section that the result is contained in. I tested in chrome and in IE and did not see any issues. The demo scripts (original mid year demo script and the new command demo script) work . Backend tests pass

danlee1025:
@nicoleng12 Try commenting out the line in `SummaryMetadata` that is fetching the treatment options to make sure that is causing the lag.

gregquinn2001:
Resolved GQ1

gregquinn2001:
rebased

gregquinn2001:
[GQ2] resolved.

karenzheung:
This looks great. I tested in chrome and ran the backend test. I tried the flux commands, searching different sections/subsections, content, etc. Also tried creating new notes and searching those which works. Nice job!

Dtphelan1:
General question: What thought have we given to search being condition specific? I lean towards being able to search everything as we do now, but this introduces a minor confusion point: sections that are condition specific don't specify the condition that the data is associated with in the search results. 

For example, typing "13 Jan" suggests that Condition has a diagnosis date of "13 Jan", but that's the diagnosis date for the fracture. Similarly, looking for "diagnosis date" shows the dates for both fracture and cancer, without providing information for which is which. 

Maybe this is a future UX improvement, but I was curious what people thought about it. If it seems critical, it may be a DP1

Dtphelan1:
Also, have we investigated the root cause of the lag mentioned by nicole?

mgramigna:
I am not seeing the same lag (weird). But to test if the API call is the cause of the lag, try commenting out the block that starts on line 2078 in SummaryMetadata.jsx. If that fixes it, I can look into improving the amount of API calls on my end regardless of whether or not I see the lag.

Dtphelan1:
Based on my local diagnostics, looks like the API call is to blame. If we have a way of doing that call at application startup and caching the results/failure, performing a check before hand to avoid multiple redundant calls we could probably avoid this issue.

mgramigna:
Yeah that makes sense. Right now we're not actually indexing the treatment options data at all, just the section name. Perhaps it makes sense to add a guard against doing the API call for data indexing for now and then make a Jira for caching the results to be able to search on?

Dtphelan1:
Yeah that sounds reasonable!

jafeltra:
[JA1] I went to /demo1 and the app is breaking now. I got the following error: Uncaught TypeError: row.map is not a function at VisualizerManager.jsx:190.

mgramigna:
I made treatment options return cached data if it is defined which stops it from making API calls after the initial one. @Dtphelan1 can you confirm that this eliminates the lag?

Looking into JA1 now

Dtphelan1:
Looks like that fixes the lag on my end. It's far from lightning fast still, but the branch no longer exhibits a major slow-down compared to master/deployment.

mgramigna:
JA1 resolved. @jafeltra brings up a good point. Testing this on all 3 patients is a good idea since they all have different data.

jafeltra:
[JA2] Switching to narrative view of the summary section for /demo1 causes the following error: Uncaught TypeError: Cannot read property '0' of undefined at _addLabResultToNarrative (NarrativeNameValuePairsVisualizer.jsx:99). Only the Summary section has this error, so it may be somewhat related to JA1 and the Recent Lab Results.

mgramigna:
JA2 resolved

mgramigna:
@daniellee1025 the issue that you brought up with the editable signed note was a simple boolean error that passed in note.signed instead of !note.signed. I think it would be fine fixing it as part of this branch?

danlee1025:
Cool I'll close out the JIRA.

PR #447, opened 2018-09-13
suppress ONCOHIST for non-cancer conditions

gregquinn2001:

Very simple change to address JIRA 1409. @ONCOHIST no longer shows up as option if fracture condition is context. If breast cancer on demo2 or GIST cancer on pilot 1, @ONCOHIST is an option.

## Submitter:

**Running the Application**

- [ ] Manually tested in IE

**Documentation**

**Code Quality**

- [X] Code is commented

**Tests**

- [X] Existing tests passed

## Reviewer 1: Name

## Reviewer 2: Name


danlee1025:
Looks good!  I also noticed that `#nlptoxicity` shows up as an option if `Fracture` condition is in context so I pushed up a commit to fix that.

gregquinn2001:
Nice. Thanks!

mgramigna:
[MG1] @ONCOHIST is no longer properly recognized as a shortcut (isn't highlighted blue) when inserting a template such as Progress Note

gregquinn2001:
MG1 is a great find! With @jafeltra help, it's "resolved". We just may end up with an unresolved @ONCOHIST shortcut in our note if they choose Fracture

jafeltra:
I think this might have to do with an edge case that Greg and I talked about the other day, but I tried testing MG1 with the 'test' template and it still works a little strangely, which I think is due to the multiple conditions and @ONCOHISTs. There are various combinations of cases that produce unexpected results, but one of the first ones I noticed was Choosing Fracture as the first condition still fills in the @ONCOHIST for Fracture within the template. If having two conditions and two @ONCOHISTs isn't a case we want to support, then we probably don't need to worry about this.

gregquinn2001:
I did notice the dual-condition weirdness. If you do cancer then fracture you can insert @ONCOHIST because of the global cancer context but it inserts using the fracture as its parent context. I think this is a JIRA.  If a shortcut requires a certain value object type for a context, then when picking the actual parent context it should find one.

In general though, we need to think about global contexts more. Maybe there should only be one active (the most recent one)... So maybe that's another JIRA.

Any other cases I didn't cover?

gregquinn2001:
added JIRA 1413 for one active global context at a time.

mgramigna:
I'm good to approve and merge this after it is confirmed to be rebased!

PR #450, opened 2018-09-14
xAxis ticks now have min, max, and 4 additionally equally spaced ticks

Dtphelan1:
Title about sums it up. We've moved the date of diagnosis to the label on the line, which makes it easier to decide which ticks to use and to ensure that as many of them display as possible. 

There is a minor issue which is that, on small screens with a note open, the min-tick along with a few others don't display because of how the component decides to automatically handle the overlap of ticks, by automatically generating them itself. Additionally, the diagnosis date label bleeds over into the y-axis on such a small screen. 

Wasn't sure if we wanted to make these JIRA's or include them in the scope of this task. Will defer to @gregquinn2001  here.

### DIFF
--- a/src/patientControl/DiseaseStatusValuesIndexer.js
+++ b/src/patientControl/DiseaseStatusValuesIndexer.js
@@ -17,7 +17,7 @@ class DiseaseStatusValuesIndexer extends BaseIndexer {
                 section,
                 subsection: "",
                 valueTitle: progression.start_time,
-                value: progression.status
+                value: progression.disease_status_string
             });
         });
     }
--- a/src/patientControl/PatientSearch.jsx
+++ b/src/patientControl/PatientSearch.jsx
@@ -139,7 +139,6 @@ class PatientSearch extends React.Component {
             }
             return suggestions;
         }, []); 
-
         return suggestions;
     }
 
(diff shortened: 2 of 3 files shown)


danlee1025:
https://github.com/FluxNotes/flux/blob/65d707ee188d82ca402a557389eb0625bcf73344/src/summary/ProgressionLineChartVisualizer.jsx#L154-L163

[DL1]  Missing semicolons in some of these lines.  Also, not a big deal, but I think `domainLength` and `tickLength` can be pulled out of the `for` loop so that they aren't calculated each time.

Dtphelan1:
Addressed DL1. 

@gregquinn2001 It's not great on a specially small screen. I can try and change the display depending on how much real estate we have, but no promises. 👍

Dtphelan1:
I've improved the display of the ticks on small screens, but we still need to revisit the diagnosis date label on small screens. Not a direct part of this sub-task, but it was done as a part of the greater task. Not sure how we want to manage that, but again -- this is definitely an improvement.

Dtphelan1:
@gregquinn2001 Let me know what you think!

karenzheung:
Removing status in ProgressionLineChartVisualizer broke the searching of dates. I talked to Greg and we made a one line change in DiseaseStatusValueIndexer to look for disease_status_String instead of status.

PR #460, opened 2018-09-21
Sorts suggestions by context, listing triggers for most recently mentioned shortcuts first

Dtphelan1:
Ensures we list the shortcuts in the suggestion portal in order of context. First step of improving the order of suggestions in the suggestion portal.


Dtphelan1:
Thanks Greg!   
  

  
  

  
  
>

PR #464, opened 2018-09-25
Sorting function improvement

Dtphelan1:
Uses fuse.js to provide an improved search function. In order to avoid re-indexing the fuse.js search indexes every time we re-render, I've introduced classes corresponding to the indexes and a the idea of 'subscribers' to the contextManager. Anything that 'subscribes' to the contextManager defines a callback for the contextManager to call whenever the contextManager updates; this callback will receive the contextManager as an argument and use that updated context a) to determine if that indexer needs to update its fuse.js index, and b) as the new context needed to perform that update appropriately. 

**N.B.** I don't love the names of the suggestionPortalSearchIndexes or where I keep them in our file structure. Would love to hear any suggestions people have on these fronts.


gregquinn2001:
[GQ1] When I type @ or # I don't get a menu immediately. We use to get that menu right after typing the special character @ or # or <. Can we get that back? I feel like that's when the user needs the most help!

gregquinn2001:
[GQ2] If I type @condition[[cancer]] #toxicity #ane 

I get these options: 
anemia
skin and other subcutaneous tissue disorders - other, specify
anorgasmia
neck edema
unenrolled
ankle fracture

It looks like a patient context shortcut (unenrolled) is sneaking into the middle of the adverse events. Since I'm in a local context, all of its recommended child shortcuts should come before any patient ones I think.

Dtphelan1:
GQ1 - Great observation I'll get right on it
GQ2 - We haven't done anything yet in the way of combining context ordering and matching score. That'd to come in a future PR. Good eyes though!

gregquinn2001:
Yeah I should have read/looked at JIRA more closely on GQ2. My bad! I retract GQ2! :)

karenzheung:
Aside from the menu not opening like greg mentioned, this looks good to me.

Dtphelan1:
Fixed GQ1. Also looked like placeholder autocomplete had gotten broken, but that's fixed as well -- no GQ necessary :)

PR #468, opened 2018-10-01
Move hover to touch functionality

danlee1025:
Nicole added functionality to show tooltip for a timeline item when user is touching the item.

### DIFF
--- a/src/timeline/TimelineEventsVisualizer.jsx
+++ b/src/timeline/TimelineEventsVisualizer.jsx
@@ -6,7 +6,6 @@ import Timeline from 'react-calendar-timeline';
 import containerResizeDetector from 'react-calendar-timeline/lib/resize-detector/container';
 import Item from './Item';
 import Button from '../elements/Button';
-import Portal from 'react-portal';
 import moment from 'moment';
 import './Timeline.css';
 import './TimelineEventsVisualizer.css';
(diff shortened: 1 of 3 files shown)


src/timeline/TimelineEventsVisualizer.scss
```diff
@@ -2,7 +2,48 @@
 
 #timeline {
   margin: 20px;
-  text-align: left;
+  text-align: left;
```
jafeltra: [JA1]: I don't think we need any of this styling anymore since we're sticking with our HoverItem for this approach.

jafeltra:
I think once we take out the old css, this should be good!

src/timeline/TimelineEventsVisualizer.jsx
```diff
@@ -275,12 +291,24 @@ class TimelineEventsVisualizer extends Component {
                         lineHeight={40}
                         itemHeightRatio={0.7}
                         itemRenderer={Item}
+                        itemTouchSendsClick={false}
```
jafeltra: [JA2]: I also don't think we're using this prop anymore since false is also the default. But if that's wrong, let me know!

danlee1025:
JA1 and JA2 should be fixed!

jafeltra:
Looks good!

PR #470, opened 2018-10-01
Merge context and matching score

Dtphelan1:
Uses the associated parent context of a shortcut suggestion in determining the suggestion's score. Specifically, the more recent the context, the smaller the score added (since the suggestions with the lowest score are listed highest in the suggestions).

Also gives the suggestionPortalPlaceholderIndex its own array of values to suggest separate from the array that the Shortcut index uses.

### DIFF
--- a/src/context/ContextManager.jsx
+++ b/src/context/ContextManager.jsx
@@ -63,39 +63,36 @@ class ContextManager {
         }, [])
     }
 
-    // Returns all currently valid shortcuts, ordering them from the most recently active context down to the patient context 
+    // Returns objects corresponding to all currently valid shortcuts, 
+    // Ordering them from the most recently active context down to the patient context 
+    // ShortcutObjects have an 'id': a string associated to their unique shortcut id
+    // and they have a 'parentId': a string associated with their parent context, if they have one other than patientContext
     getCurrentlyValidShortcuts(shortcutManager) {
         let result = [];
-        let childResults = [];
+        let childIds = [];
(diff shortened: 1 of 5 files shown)


jafeltra:
I tested this by adding a condition and toxicity and started adding "disease" and saw that toxicity adverse events came before the "disease status" shortcut, which also comes up before the patient level "deceased" shortcut. The code looks good to me, too. Unless there's anything else specific I should test, this should be set!

jafeltra:
I do have one question after talking to @gregquinn2001 about the expected behavior of global and local contexts. We were saying that global contexts would just prioritize based on how well the text matches and not by what is in the most recent context, but I don't see that happening. If I have GIST added as a context, if I type "#u", both "disease status" and "unenrolled" match on the letter u. However, since it's a global context and "unenrolled" actually starts with the letter you typed, I think it is expected that "unenrolled" would be first. Right now, "disease status" is the first option. Which order is the expected order? If it's not what is currently implemented and this should be a Jira, that works too.

Dtphelan1:
Global contexts don't have special treatment in the current implementation; they are treated like any other context.

Dtphelan1:
I've come up with a way of treating global contexts the same as parent contexts. Let me know if this works! 👍

jafeltra:
Awesome! This looks good to me! The prioritization of suggestions seems really intuitive to me!

jafeltra:
This just needs to be rebased again.

PR #474, opened 2018-10-04
Improve open note matching

mgramigna:

For Jira 1453. With Fuse.js, the searching of open note content was behaving strangely. I changed it to use lunrjs now, and the behavior seems to be more like what you would expect.

Compare with searching open note content on fluxnotes.org, e.g. Search for "Stable" or "Disease Status" and this will highlight some of the improvements for this PR.

## Submitter:

**Running the Application**

- [ ] Manually tested in IE

**Documentation**

- [ ] Cheat sheet is updated
- [ ] Demo script is updated 

**NoteParser**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "history": "^4.7.2",
     "immutable": "3.8.1",
     "interact.js": "1.2.8",
+    "lunr": "^2.3.3",
     "material-ui": "v1.0.0-beta.30",
     "material-ui-icons": "1.0.0-beta.15",
     "mkdirp": "0.5.1",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "history": "^4.7.2",
     "immutable": "3.8.1",
     "interact.js": "1.2.8",
-    "lunr": "^2.3.3",
     "material-ui": "v1.0.0-beta.30",
     "material-ui-icons": "1.0.0-beta.15",
     "mkdirp": "0.5.1",
--- a/src/notes/FluxNotesEditor.jsx
+++ b/src/notes/FluxNotesEditor.jsx
@@ -767,6 +767,7 @@ class FluxNotesEditor extends React.Component {
     }
 
     onOpenNoteSearchResultHighlight = (suggestion) => {
+        if (!suggestion.indices) return;
         const {document} = this.state.state;
         let startIndex = suggestion.indices[0];
         let foundNode;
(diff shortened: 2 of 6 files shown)


nicoleng12:
I noticed that when searching an open note, the search results still show the hashtag for the shortcut even though it's no longer displayed in the editor. Since it's not related to this PR, should we just make a jira to make the search result consistent with the content of the note?
image

nicoleng12:
[NN1] - This is small, but can we add lunrjs to the list of third party libraries in our wiki? https://github.com/FluxNotes/flux/wiki/About-Flux-Notes

mgramigna:
Resvoled NN1, good catch Nicole. I'm going to check to see if your first comment is something trivial to change and if not then a Jira probably makes sense

nicoleng12:
Overall this looks good! 

[NN2] - I noticed that you have to type the whole word in order for the search to find it in the open note. For example if the note contains toxicity, I have to write "toxicity" for the search to display the open note as a result. I would expect it to start including the open note as a search result when I type "tox" at least.

mgramigna:
Removed hashtags from search results

mgramigna:
Addressed NN2

danlee1025:
image

[DL1]  The second search results seems to highlight a lot of information irrelevant to the query.  I'm guessing it matched because of `tumor` but it probably shouldn't highlight the text beyond that.  What do you think of that behavior?

danlee1025:
image


[DL2] The 1st search result seems weird that it would match and it seems to match twice on the title of the open note.

mgramigna:
DL2 resolved. Investigating DL1

mgramigna:
Important update: after talking with Greg, it seems like the actual best approach for this is just using a regex. So I switched to that. Now I think it behaves how a clinician would expect it to which is the ultimate goal

mgramigna:
rebased

danlee1025:
[DL3]  When I open the 1st note and search for `abdomin` the correct search results show up, but when I search for `abdomina` there are no more search results.  Then when I search for `abdominal` the search results appear again.  Another example is `toxic` -> `toxici` -> `toxicity`.

image

image

mgramigna:
DL3 is a result of elasticlunr tokenizing the strings weirdly. Looking into it to see if there's an easy solution but if not I think it might warrant a Jira since it's unrelated to open note matching

mgramigna:
Made a Jira for DL3

nicoleng12:
Looks good! Nice job Matt!

PR #479, opened 2018-10-10
Initial fix of lag

Dtphelan1:
Store the result of our parsing in the data_cache, and check to see if the data_cache has data we can use before trying to load the promise data; if that data already exists, just return.


danlee1025:
Looks good, seems to fix the issue of continuously requesting from the server.  Just pushed a commit to add some semicolons..

Dtphelan1:
How could I ever doubt you @jafeltra?

PR #490, opened 2018-10-22
Add json file to set server url and Treatment Option section creates …

karenzheung:
…api using config settings from file

JIRA 1326

Add new file ServerConfiguration.json on root directory of project where user can change baseURL that the treatment options api makes call to. 

## Submitter:

**Running the Application**

- [ ] Manually tested in IE

**Documentation**

- [ ] Cheat sheet is updated
- [ ] Demo script is updated 

**NoteParser**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/ServerConfig.json
+++ b/ServerConfig.json
@@ -1,3 +1,0 @@
-{
-    "baseURL": "http://localhost:3002/api"
-}
--- a/ServerConfig.json
+++ b/ServerConfig.json
@@ -1,0 +1,3 @@
+{
+    "baseURL": "http://localhost:3002/api"
+}
--- a/public/ServerConfig.json
+++ b/public/ServerConfig.json
@@ -1,0 +1,3 @@
+{
+    "baseURL": "http://localhost:3001/api"
+}
--- a/src/summary/metadata/TreatmentOptionsSection.jsx
+++ b/src/summary/metadata/TreatmentOptionsSection.jsx
@@ -4,12 +4,6 @@ import Lang from 'lodash'
 //import config from '../../../ServerConfig.json';
 const ApiClient = new FluxNotesTreatmentOptionsRestClient.ApiClient();
 
-
-fetch('/ServerConfig.json').then((r) => r.json()).then((config) => {
-    ApiClient.basePath = config.baseURL;
-})
-
-const api = new FluxNotesTreatmentOptionsRestClient.DefaultApi(ApiClient);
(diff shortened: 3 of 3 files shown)


jafeltra:
Hi Karen! Are there steps I can follow to confirm that the server is using the configured base URL? Should I also be looking for changes in the node server repository? I did try changing the port in the server repository's index.js file and I'm getting errors trying to load parsedData. Any suggestions?

karenzheung:
@jafeltra There are no changes to the node server or api client repository. When you load the page, if inspect and go to network, you should see that the GET request is making a request using http://localhost:3002 instead of the normal port 3001. If you go to the ServerConfiguation.json file in the root flux/ and change the baseURL to something else, you should see the new url in GET request on network tab. Currently server will not return data because the server location hasn't been changed. I don't think the was in the scope of the JIRA. But I think @gregquinn2001  should confirm that.

Dtphelan1:
Hey hey! Two questions/possible issues: 
1. Is this server config file actually editable in our build folder? It looks like, even if it is, it's pretty buried in our directories. Do we want to pull it into the `public` folder so that, on build, it will be in the base level directory?
2. Is there really no better way to edit basePath in swagger from the client side? It feels really risky to me that we are making assumptions about the member variables that the ApiClient exposes. If that's really the limits of swagger, than I guess that's fine; but i'd feel much more secure calling a function that we can depend on to change the basePath.

jafeltra:
Thanks for the clarification Karen! Since we don't need to update the server repo yet, I think this looks good, besides the two questions that Dylan had. I think before we merge this, we should change the configured URL to be 3001 so that it does work with the server when it is running. Do you agree?

gregquinn2001:
@jafeltra  and @karenzheung . Let's change the URL back to the one that works (port 3001). Can we also move the ServerConfig.json file to the public folder. We can then test whether after build and deploy on the shr-dev server that we can edit the ServiceConfig.json file, restart the web server (apache I believe), and then the treatment options server url will be the new one we put in the file.

jafeltra:
[JA1] I tried to change the serverConfig.json file locally to use port 3002, but it seems like there might be a timing issue because I can still connect to the treatment options server when it is running on 3001. Should the server crash if I've changed the url?

gregquinn2001:
@jafeltra and I resolved JA1

PR #514, opened 2018-11-08
Improved note template

nicoleng12:
This PR addresses jira 1499. 

- Removed "progress notes" template. Spoke with the team and it was decided that instead of renaming the template, just remove it since it's not being used. Also remove "op note" template because that was also not used

- Edited "Follow up" and "Initial consult" templates to include date of visit and referring physician

- Created new shortcuts to support adding new information to the templates. Created @date of visit and @referred by inserters

- Updated patient json files for Ihannos, Ella, and Debra to include ReferredBy

-  Spoke with @gregquinn2001 about including a header which pulls in age and gender. Since this information is already displayed in the visit reason section, we decided not to add an additional section and wait for further clinician feedback

## Submitter:

**Running the Application**

- [x] Manually tested in IE

**Documentation**

- N/A Cheat sheet is updated
- N/A Demo script is updated 

**NoteParser**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewer 1: Name

## Reviewer 2: Name

### DIFF
--- a/src/context/TemplateViewModeContent.jsx
+++ b/src/context/TemplateViewModeContent.jsx
@@ -9,8 +9,7 @@ export default class TemplateViewModeContent extends Component {
         { name: 'physical exam', content: 'Vitals: @vitals\nSkin:\nLymph:' },
         { name: 'follow-up', content: 'REASON FOR VISIT:\n@patient following up on @condition on @date of visit. @reason for next visit.\n\nONCOLOGY HISTORY:\n@ONCOHIST\n\nHISTORY OF PRESENT ILLNESS:\n\n\nREVIEW OF SYSTEMS:\n@ROS\n\nALLERGIES:\n@ALLERGIES\n\nMEDICATIONS:\n@active medications\n\nPHYSICAL EXAM:\n\n\nASSESSMENT:\n\n\nPLAN:\n\n' },
         { name: 'initial consult', content: 'REASON FOR VISIT:\n@patient presenting with @condition for initial consult on @date of visit. @reason for next visit. @referred by.\n\nHISTORY OF PRESENT ILLNESS:\n\n\nREVIEW OF SYSTEMS:\n@ROS\n\nALLERGIES:\n@ALLERGIES\n\nMEDICATIONS:\n@active medications\n\nPHYSICAL EXAM:\n\n\nASSESSMENT:\n\n\nPLAN:\n\n' },
-        { name: 'test', content: 'FOLLOW UP:\nPatient is showing signs of @condition @ONCOHIST @condition @ONCOHIST\n\nMEDICATIONS:\n@medication\n\nProcedures:\n@procedure' },
-        { name: 'consult note', content: '@patient presenting with ' }
+        { name: 'test', content: 'FOLLOW UP:\nPatient is showing signs of @condition @ONCOHIST @condition @ONCOHIST\n\nMEDICATIONS:\n@medication\n\nProcedures:\n@procedure' }
     ];
 
     // Insert the content of the template as you would a shortcut
--- a/src/context/TemplateViewModeContent.jsx
+++ b/src/context/TemplateViewModeContent.jsx
@@ -7,8 +7,8 @@ export default class TemplateViewModeContent extends Component {
     // These don't need to be stored as state since they're not being updated
     templates = [
         { name: 'physical exam', content: 'Vitals: @vitals\nSkin:\nLymph:' },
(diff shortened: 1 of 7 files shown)


gregquinn2001:
I didn't look but I just read your pull request and it made me think a little more. I should have mentioned this before, but I'm assuming you added a referred by field to the most recent encounters. For pilot 1, I only think the first encounter with the medical oncologist would have a referred by from the PCP of the first note (so no new names). All the other encounters wouldn't have a referred by because they are follow-ups with their oncologist. Make sense? Sorry! If there is no referred by on the current appointment, the note text should not have any content also. That case becomes more important since it's the case for the pilot1 note being written during our usability testing. Demo 2 and Demo 1 are not as important but demo 2 ideally would also have referrals only where they make sense (which may be none at all)

nicoleng12:
@gregquinn2001 I made the changes we discussed. I removed @referred by from the follow-up note template and added text to indicate no referral when there is no referral

gregquinn2001:
[GQ2] @Date of visit is showing a date of 22 Sept 2018 which is date of previous encounter - not today's. I think this was supposed to be the encounter we're writing the note about so the most recent or today's. The problem is that it hasn't happened yet. Today's encounter or the most recent one maybe should be used instead of most recent one in past?

nicoleng12:
Rebased and addressed GQ2

gregquinn2001:
Hey Nicole the method you added should be called getTodayOrMostRecentEncounterDate or it may be better to call a method such as getTodaysEncounter and if that returns null call the current getMostRecentEncounterDate method. The method you added does something different than it name indicates IMO.

nicoleng12:
@gregquinn2001 I renamed the method to getTodaysEncounter

nicoleng12:
Rebased

nicoleng12:
@Dtphelan1 thanks for cleaning up the code! This looks good to me if you want to approve and  merge

jafeltra:
[JA1]: I pulled this branch to double check it before merging since I think Dylan is out today and I got this warning when inserting the follow up template:

moment.js:293 Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: 14 Nov 2018, _f: undefined, _strict: undefined, _locale: [object Object]
Error
    at Function.createFromInputFallback (http://localhost:3000/static/js/bundle.js:187794:99)
    at configFromString (http://localhost:3000/static/js/bundle.js:189842:16)
    at configFromInput (http://localhost:3000/static/js/bundle.js:190068:14)
    at prepareConfig (http://localhost:3000/static/js/bundle.js:190051:14)
    at createFromConfig (http://localhost:3000/static/js/bundle.js:190018:45)
    at createLocalOrUTC (http://localhost:3000/static/js/bundle.js:190105:17)
    at createLocal (http://localhost:3000/static/js/bundle.js:190109:17)
    at Moment.isSame (http://localhost:3000/static/js/bundle.js:190695:53)
    at http://localhost:3000/static/js/bundle.js:408348:95
    at baseFindIndex (http://localhost:3000/static/js/bundle.js:96164:12)

jafeltra:
@Dtphelan1 Thanks for the update! I did notice that if we compare the today date and the encounter date using `===` instead of `isSame()`, the warning goes away. So like 314 in PatientRecord becomes `return new moment(encounter.expectedPerformanceTime, "D MMM YYYY") === today;` instead of `return new moment(encounter.expectedPerformanceTime, "D MMM YYYY").isSame(today, 'day')`. But I also noticed you changed that in your last commit. Is that not enough of a comparison for some reason?

Dtphelan1:
Great question.   
  
Short: it’s probably fine to use the old way for now to suppress the warning.   
  

  
Long: The idiomatic way to do comparisons in Moment is using the isSame fn seen there. Ultimately we have always been using a format Moment isn’t thrilled with; we just haven’t encountered the warning until now. That’s my assessment at least. The format fn is probably pretty flexible, so Moment hasn’t yelled at us yet; but those object oriented fns are probably way more explicit with their warnsings and checks. So if we were building something for production and wanted to use Moment as “by-the-books” as possible we’d probably want to use the member functions; but we’re not, and we haven’t yet, so it’s probably fine to do the regular old str comparison. :)
  

  

  
  

  
  
>

nicoleng12:
Great comments everyone! @Dtphelan1 we will go back to using the regular old str comparison to avoid warnings for now, but you make a really great point. I'm going to make a jira for going back through our code to make sure we are using member functions for Moment js.

PR #520, opened 2018-11-14
Fixed app crash when there is no recent staging value.

danlee1025:

`/demo1/` was crashing because of missing stage value.


sgober:
[SO1] One of the tests doesn't pass

danlee1025:
Good catch Sarah!  Fixed the broken test.

PR #557, opened 2019-01-31
Asynch patient fetching

Dtphelan1:
This  PR adds in the functionality to support asynchronous loading of patient data from a data source. In order to do so, a few main changes had to be made: 
- Main dashboards are only ever loaded when there is a non-empty patient in the local state. This avoids having to make the dashboards sophisticated enough to handle conditionally interacting with and using patient data. 
- The PatientSummaryHeader, on the other hand, _was_ changed to handle having no patient data. In this case, it only renders the logo and the current user information. 
- We can now display loadingAnimations and loadingErrors. These are always rendered, but only fade into the display when loading is happening in the first case and when an error occured in loading patient information in the second. 
- Added 'Gestalt' information to each of our data sources, describing what operations are supported and what kinds of requestTypes are supported.  On the origins of this term see [here](https://en.wikipedia.org/wiki/Gestalt_(Mac_OS). For each dataSource, this defines what kind of requests can be made and whether or not the request is synchronous or asynchronous. At the app level, we use that information to determine whether or not to send a callback along to get the patient data. 
- **Need to `yarn install`** since I updated our version of material ui from `1.0.0-beta.30` to `10.0.-beta.32`. We needed this upgrade to fix a bug in the `Fade` component  that gets used to fade in and out the application on loading. This was the absolute minimum requirement necessary, as evidenced by the changelog found [here](https://github.com/mui-org/material-ui/blob/master/CHANGELOG.md#component-fixes--enhancements-33). 
    - **N.B.** This introduced a new breaking change related to the `Button` component that I have since fixed, described in the list [here](https://github.com/mui-org/material-ui/blob/master/CHANGELOG.md#breaking-change-32).

For Testing: 
- Visit `pilot1` to confirm that synchronous loading of hardcoded data still works (timeout added for effect) 
- Visit `ccp` to confirm that synchronous loading of hardcoded data still works (no timeout)
- Visit `p2` with the local node server up to confirm that synchronous loading of remote data still works. 
- Visit `p2` without the local node server up to confirm that errors are displayed well in the sync case. 
- Visit `p2` with the local node server up, AND modify the local API file in `node_modules/shr_rest_client/src/api/DefaultApi.js` to use the asynchronous fetch to confirm that errors are displayed well in the asynch case. 
- We obviously can't confirm that async fetch itself works flawlessly until we have such a resource

### NOTE: I've left in `setTimeout`'s throughout the loading of patient data so that reviewers can judge the look/feel of the loading animations. These **need to be removed** before merging. 

## Submitter:

**Running the Application**

**Task**

**Documentation**

- N/A Pre release script is updated on the [Pre Release Testing Script wiki page](https://github.com/FluxNotes/flux/wiki/Pre-Release-Testing-Script) (add/remove single features to test, do not run through full script)
- N/A Update the list of key features with new features on the [Key Application Features wiki page](https://github.com/FluxNotes/flux/wiki/Key-Application-Features)
- N/A Document any new APIs/extension points
- N/A Additional technical components and libraries are documented and added to [wiki](https://github.com/FluxNotes/flux/wiki/About-Flux-Notes) as needed

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/containers/FullApp.jsx
+++ b/src/containers/FullApp.jsx
@@ -59,6 +59,7 @@ export class FullApp extends Component {
             "encounter",
             "post-encounter"
         ];
+        // Determines how long the fade-in v. fade-out animation lasts
         this.timeoutDuration = 1000;
 
         if (Lang.isUndefined(this.props.dataSource)) {
@@ -88,6 +89,8 @@ export class FullApp extends Component {
             layout: "",
             // Start the app loading information
             loading: true,
+            // If there is an error produced when loading data, it will go here
+            loadingError: null,
             loginUser: {},
             modalTitle: '',
             modalContent: '',
@@ -152,17 +155,31 @@ export class FullApp extends Component {
     }
 
     loadPatient(patientId) {
-        let patient = this.dataAccess.getPatient(patientId);
-        this.contextManager = new ContextManager(patient, this.onContextUpdate);
-        if (patient) { 
-            this.setState({ 
-                patient, 
-                loading: false
-            })
+        if (this.dataAccess.getGestalt().requestTypes.async) { 
+            console.log('async')
+            this.dataAccess.getPatient(patientId, (patient, error) => { 
+                this.contextManager = new ContextManager(patient, this.onContextUpdate);
+                setTimeout(()=>this.setState({ 
+                    patient, 
+                    loading: false,
+                    loadingError: error
+                }), 3000);
+            });
         } else { 
-            this.setState({ 
-                loading: false
-            })
+            // Assume sync
+            try {
+                let patient = this.dataAccess.getPatient(patientId);
+                this.contextManager = new ContextManager(patient, this.onContextUpdate);
+                this.setState({ 
+                    patient, 
+                    loading: false
+                });
+            } catch (error) {
+                this.setState({
+                    loading: false, 
+                    loadingError: error
+                })
+            }
         }
     }
 
@@ -177,12 +194,8 @@ export class FullApp extends Component {
     }
 
     componentDidMount() { 
-        // Add a non-burdensome timeout so that the loading animation doesn't flicker oppreseively
-        //TODO: REMOVE -- FOR TESTING ANIMATION
-        setTimeout( () => 
-            this.loadPatient(this.props.patientId),
-        2000)
-        // TESTING
+        // Once the component has mounted, we can try to load the patient data
+        this.loadPatient(this.props.patientId)
     }
 
     receive_command(commandType, data) {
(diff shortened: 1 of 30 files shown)


src/dataaccess/IDataSource.jsx
```diff
@@ -13,6 +13,23 @@ class IDataSource {
...
         throw new Error("savePatient not implemented by " + this.constructor.name);
     }
+
+    // On the origins of this term see: https://en.wikipedia.org/wiki/Gestalt_(Mac_OS)
+    getGestalt() { 
+        // 
+        // Shape of the gestalt : { 
+        //      requestTypes: {
+        //          sync: bool
+        //          asynch: bool
```
dehall: Quick note that the comment here says `asynch` but everywhere else in the code it's `async`.

src/dataaccess/DataAccess.jsx
```diff
@@ -36,4 +36,8 @@ export default class DataAccess {
     savePatient(patient) {
```
dehall: I understand that it's not part of this task, but at some point the other functions should probably be updated to support sync/async as well. As a heads-up the SMART on FHIR client library makes all service calls asynchronously, so the savePatient function in that upcoming data source will have to be async.

Dtphelan1: Agreed. I can make a JIRA to support thos additional functions. Thanks Dylan!

src/dataaccess/IDataSource.jsx
Dtphelan1: Taken care of! 👍

src/containers/FullApp.jsx
```diff
@@ -148,13 +157,56 @@ export class FullApp extends Component {
...
+                    patient, 
+                    loading: false,
+                }), 5000);
+                //
+                // TODO: RESTORE AFTER TESTING
+                // this.setState({
+                //     patient, 
+                //     loading: false
+                // });
+            } catch (error) {
```
dehall: If there is an error somewhere, nothing from here on gives you easy access to the full stack trace, unless you're already debugging. Same with the async path above. The LoadingError object just shows the message which isn't always sufficient.
I'm not sure if there's a Flux standard for error logging, but it could be helpful to write the error to console.error either here or in the LoadingError class, that way all the individual DataSource classes don't necessarily have to wrap everything in try/catch with a logger. Compare to before when there was no error handler and any errors would just bubble up.

Dtphelan1: I like it -- implementing now.

Dtphelan1: Nice. I added that functionality into the different apps, so that's logged properly as errors.

src/dataaccess/RestApiDataSource.jsx
```diff
@@ -6,14 +6,32 @@ class RestApiDataSource extends IDataSource {
...
-        // return this.api.getPatientById(id, callback);
-        
+
+    getGestalt() { 
+        return this._gestalt
+    }
+
+    getPatient(id, callback) {
+        // Asynchronous call
+        // return this.api.getPatientById(id, (error, data, response) => {
```
dehall: Should this still be commented? In the description of the task it says:
> This will also involve changing the shr-javascript-client library to remove the additional code to make a synchronous call.
> When this is working, the commented out test in DataAccess.test.js for getting a patient from the RestApiDataSource should be uncommented and should pass.

So unless something changed I think this async version should be uncommented and that test case should be uncommented and changed to use the callback method. To correctly test an async call it should look something like this where it calls `done` in the callback:
```
    it('should do a thing async', function(done) {
        dataAccess.getPatient(id, (result, error) => {
            expect(...);
            done();
        });
    });
```

Dtphelan1: Great observation. From the discussions I had with @gregquinn2001, we would have liked to have gotten the RestApiDataSource working but we didn't want to hold up this task if we couldn't. In this case, I made changes to the shr_rest_client for testing but ran into a litany of CORS issues that were relentless. Because of that, I decided to punt on that for now, particularly because our next focus should be on getting @dehall's new data source up and running and not this older data source.

tl;dr Tried to fix this async; ran into CORS issues; punting because this dataSource isn't a priority

Dtphelan1:
Bump on @gregquinn2001  👍  Taking accountability on my PR 😄

Dtphelan1:
Tried a new shape to the gestalt -- take a peek! 👍

gregquinn2001:
Looks good. 1 thought though. We don't need isSupported really right? if async and sync or both false then the operation isn't supported at all? Too subtle?

gregquinn2001:
needs a rebase?

Dtphelan1:
Rebasing and removing the isSupported flag 👍

Dtphelan1:
All set on the rebase. Removing the testing timeouts and then I think we should be good to merge

PR #566, opened 2019-02-07
Copy paste

karenzheung:

## Submitter:
JIRA 1618

Fixes the copy and paste bug with zero-width space on structured fields by eliminating the use of void nodes. 

**Running the Application**

**Task**

**Documentation**

- [x] Pre release script is updated on the [Pre Release Testing Script wiki page](https://github.com/FluxNotes/flux/wiki/Pre-Release-Testing-Script) (add/remove single features to test, do not run through full script)
- [x] Update the list of key features with new features on the [Key Application Features wiki page](https://github.com/FluxNotes/flux/wiki/Key-Application-Features)
- [x] Document any new APIs/extension points

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
+++ b/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
@@ -53,21 +53,17 @@ function SingleHashtagKeywordStructuredFieldPlugin(opts) {
                 // Sort keywords based on length -- we want to match longest options first
                 keywords.sort(_sortKeywordByNameLength);
                 const keywordInClosetBlock = scanTextForKeywordObject(curNode, keywords)
-
                 if (!Lang.isUndefined(keywordInClosetBlock)) {
                     const keywordText = keywordInClosetBlock.name.toLowerCase();
                     const newKeywordShortcut = createShortcut(null, keywordText);
                     newKeywordShortcut.setSource("Keyword");
-             
 
                     // KeywordRange never be null -- we've already confirmed the existance of the keyword
                     let keywordRange;
                     if (curNode.nodes) {
                         for (const childNode of curNode.nodes) {
-                           
                                 keywordRange = getRangeForKeyword(childNode, keywordText);
                                 if (keywordRange) break;
-                            
                         }
                     } else {
                         keywordRange = getRangeForKeyword(curNode, keywordText);
--- a/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
+++ b/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
@@ -121,7 +121,7 @@ function SingleHashtagKeywordStructuredFieldPlugin(opts) {
         let curNodeText = [];
         if(curNode.nodes){
             for(const childNode of curNode.nodes){
-                if(childNode.type != 'structured_field')
+                if(childNode.type !== 'structured_field')
                     curNodeText.push(childNode.text);
             }
         }
--- a/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
+++ b/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
@@ -62,8 +62,8 @@ function SingleHashtagKeywordStructuredFieldPlugin(opts) {
                     let keywordRange;
                     if (curNode.nodes) {
                         for (const childNode of curNode.nodes) {
-                                keywordRange = getRangeForKeyword(childNode, keywordText);
-                                if (keywordRange) break;
+                            keywordRange = getRangeForKeyword(childNode, keywordText);
+                            if (keywordRange) break;
                         }
                     } else {
                         keywordRange = getRangeForKeyword(curNode, keywordText);
--- a/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
+++ b/src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
@@ -37,6 +37,7 @@ function SingleHashtagKeywordStructuredFieldPlugin(opts) {
         if (relevantSingleHashtagKeywordMappings.length !== 0) {
             // Get all relevant keywordShortcuts, 
             const listOfKeywordShortcutClasses = findRelevantKeywordShortcutClasses(listOfSingleHashtagKeywordShortcutMappings).reduce((accumulator, listOfKeywordsForShortcut) => accumulator.concat(listOfKeywordsForShortcut));
+         
             for (const keywordClass of listOfKeywordShortcutClasses) {
                 // Scan text to find any necessary replacements 
                 let keywords = getKeywordsBasedOnShortcutClass(keywordClass);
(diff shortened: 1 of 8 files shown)


jafeltra:
[JA3] If I highlight starting directly next to a shortcut (I think I am _not_ grabbing the zero-width-space) and copy and paste the text, it does not paste as structured data. Do we want it to?

[JA2] I'm getting a few warnings in the console about unused variables and an `expected !== but saw !=` warning.

[JA3] There seems to be a lot of whitespace changes that we can get rid of. For example, in the `by-key.js` file, we can probably get rid of the full diff we we delete the whitespace changes made. There are other lines that have extra spaces (like various lines in SingleHashtagKeywordStructuredFieldPlugin from lines 61 - 70). Ideally, I think new lines should not have any spaces or tabs in them. I can point out more specifics if you need them and if it's helpful.

jafeltra:
[JA4] Using templates has a two bugs.
1) When you hover over an option in the picklist, the text displays twice in the note. 
2) Choosing fracture from the picklist options gives an error in the console:
`Uncaught TypeError: Cannot read property 'length' of null
    at FluxNotesEditor.resetShortcutData (FluxNotesEditor.jsx:586)`
[JA5] I can get to a state when I lose my cursor. I'm not sure if this is related to changing how we render the shortcuts, but I can't seem to get it to happen on fluxnotes.org. If I insert condition as the first thing in the note, then arrow to the left, I lose the cursor. On fluxnotes.org, it appears on the left side of the shortcut after 2 arrow clicks. I also can't ever get the cursor to come back after hitting right arrow. When I am in this state and can't see the cursor, I also can't type anything.

mgramigna:
[MG1] These are more questions than change requests, I am just looking for more insight so I understand the code changes more.

1) What is the reasoning for the Slate changes in `apply-operation.js` and `operation.js`?
2) What is the reasoning for the changes in the `resetShortcutData` function in `FluxNotesEditor.jsx`?

I'm just having a hard time following, and I think we should be careful about making changes to Slate's code so I just want to understand the reasoning.

karenzheung:
Fixed all the JAs. There is an issue where paste doesn't work if the zero width space is not selected. 
Regarding MG1, I made those changes because insertText and removeText expects a text node, so when a block or inline is passed I get an error. I found this issue regarding that: 
https://github.com/ianstormtaylor/slate/issues/775

The changes in resetShortcutData were made because when inserting from the picklist, it didn't properly update the shortcut with the new selected value since the rendering of shortcuts were changed. If I selected @condition, then selected Fracture from the picklist the editor would show @conditionFracture vs Fracture.

mgramigna:
[MG2] The arrow key navigation works for hitting the left arrow, but hitting the right arrow at the beginning of an inserter causes it to jump to the end when we expect it to go character by character (works fine for hashtags, just need to keep the old arrow key functionality in for inserters)

Perhaps we should add a case to the if statement for the arrow keys to make sure that the shortcut we care about is not an Inserter before doing the custom arrow key navigation?

karenzheung:
Fixed MG2

src/notes/SingleHashtagKeywordStructuredFieldPlugin.jsx
```diff
@@ -61,8 +62,8 @@ function SingleHashtagKeywordStructuredFieldPlugin(opts) {
                     let keywordRange;
                     if (curNode.nodes) {
                         for (const childNode of curNode.nodes) {
-                            keywordRange = getRangeForKeyword(childNode, keywordText);
-                            if (keywordRange) break;
+                                keywordRange = getRangeForKeyword(childNode, keywordText);
```
mgramigna: Unnecessary whitespace (lines 65-66)

src/notes/StructuredFieldPlugin.jsx
```diff
@@ -51,13 +51,43 @@ function StructuredFieldPlugin(opts) {
...
+            if (previousNode.type === 'structured_field'
+                && !(previousNode.data.get('shortcut') instanceof InsertValue)
+                && state.selection.anchorOffset === 0 && state.selection.isCollapsed) {
+                let transform = state.transform();
+                transform = transform.removeNodeByKey(previousNode.key);
+                let newState = transform.apply();
+                return newState;
+            }
+        } 
+        else if (e.keyCode === 37 && previousNode){
```
mgramigna: Unnecessary single space (lines 73 and 81). Should probably just move the `else if (...)` up a line to be on the same line as the bracket to stay consistent with our JS style

mgramigna: Also need a space on line 74 before the bracket

src/notes/StructuredFieldPlugin.jsx
```diff
@@ -530,6 +560,7 @@ function StructuredFieldPlugin(opts) {
         utils: {
             //isSelectionInStructuredField
             convertSlateNodesToText: convertSlateNodesToText,
+            createStructuredField: createStructuredField.bind(null, opts)
```
mgramigna: Not sure we need this line at all

src/shortcuts/InsertValue.jsx
```diff
@@ -190,6 +190,7 @@ export default class InsertValue extends Shortcut {
     }
 
     getText() {
+
```
mgramigna: this can be removed

mgramigna:
[MG4] We should be updating all the schemas (except placeholders) to reflect the new non-void hashtags.

i.e. update the schema for `structured_field_selected_search_result` and `structured_field_search_result` to also get rid of `shortcut.getText()` in the rendering, similar to what was done for `structured_field` and `bolded_structured_field`

karenzheung:
Fixed MG 3 and MG4

src/notes/StructuredFieldPlugin.jsx
mgramigna: Not sure if the change for this got pushed, or do you still think we need this line?

mgramigna:
I think this looks good. I will approve pending follow up on [this comment](https://github.com/FluxNotes/flux/pull/566#discussion_r262658120)

jafeltra:
This looks good to me! All the previous JA's are resolved, pending the one last comment from Matt. I did notice one new thing that I think is related to the changes in SingleHashtagKeywordStructuredFieldPlugin.js:

[JA6]: If you type "#disease status #toxicity nausea disease " the first disease word gets captured as the structured phrase after you hit space and the second disease does not. I know this is a bit of a strange use case, but since we have a fix for it you type #disease status after #toxicity, I feel like we should support this case too.

karenzheung:
Fixed JA6

src/notes/StructuredFieldPlugin.jsx
karenzheung: I forgot this push this. pushed now

jafeltra:
JA6 looks good to me! I haven't noticed anything new so everything looks set to me! Nice job!!

PR #568, opened 2019-02-08
[CCP] Implemented separated outcomes tables with select popups

noranda:
**This PR implements the separated outcomes tables with select popups as designed in the [MVP](https://www.figma.com/proto/rLpubNAfrqV99exbLVFz5c/core-cancer-pilot?node-id=555%3A6877&viewport=1100%2C773%2C1&scaling=min-zoom) with the following exceptions to be implemented in future tasks:**
- table/icons toggle
- icons visualization
- side effects - all column
- side effects - header dropdown
- side effects - more (...) button
- legend
- other deaths in bar charts

**This PR does the following:**
- increased max memory to handle large mcode datasets
- changed the ccp patient stage from `IA` to `IIA`
- implemented new structure for outcomes table:
  - added `TreatmentsPopover` component to handle treatment selections
  - added `isSame` and `getCombinations` array utils to help with logic regarding which treatment rows to render
  - updated treatment names object
  - updated `includedTreatments` and `combinedTreatments` default values
  - added event listeners to be able to close poppers by clicking outside of them
  - wrote logic to handle various selections of treatments -- for example:
    - there should be no duplicate rows
    - rows will only show up if there is at least 1 patient with the specified treatment combination
    - if no patients exist to create a row for the included table, helper text will appear -- for the compared row, nothing will appear
- updated `BarChart` component to take in a `comparedRow` and defaulted `active` to false for the time being -- this will be updated in a future task when `BarChart` will be restructured to deal with 'other deaths'
- hooked up similar patient checkboxes for `pathology` and `treatment history`
- styled all components
- fixed and wrote new tests for `TreatmentOptionsOutcome` component
- updated the redux store dev tools extension to get rid of console deprecation warning

****Note: When testing this branch, unselect all similar patient criteria to populate the table, then check off hooked up boxes.**

**All backend and mcode tests pass. Tested in Chrome and Safari.**

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "watch-css": "npm run build-css && node-sass-chokidar src/ -o src/ --watch --recursive",
     "start-js": "react-scripts start",
     "check-version": "node ./scripts/check-version.js",
-    "start": "npm-run-all -p watch-css start-js check-version",
+    "start": "NODE_OPTIONS=--max-old-space-size=4096 npm-run-all -p watch-css start-js check-version",
     "build-js": "node --max_old_space_size=4096 node_modules/react-scripts/scripts/build.js",
     "build": "npm-run-all build-css build-js",
     "eject": "react-scripts eject",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "watch-css": "npm run build-css && node-sass-chokidar src/ -o src/ --watch --recursive",
     "start-js": "react-scripts start",
     "check-version": "node ./scripts/check-version.js",
-    "start": "NODE_OPTIONS=--max-old-space-size=4096 npm-run-all -p watch-css start-js check-version",
+    "start": "NODE_OPTIONS=--max-old-space-size=5120 npm-run-all -p watch-css start-js check-version",
     "build-js": "node --max_old_space_size=4096 node_modules/react-scripts/scripts/build.js",
     "build": "npm-run-all build-css build-js",
     "eject": "react-scripts eject",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "watch-css": "npm run build-css && node-sass-chokidar src/ -o src/ --watch --recursive",
     "start-js": "react-scripts start",
     "check-version": "node ./scripts/check-version.js",
-    "start": "NODE_OPTIONS=--max-old-space-size=5120 npm-run-all -p watch-css start-js check-version",
+    "start": "npm-run-all -p watch-css start-js check-version",
     "build-js": "node --max_old_space_size=4096 node_modules/react-scripts/scripts/build.js",
     "build": "npm-run-all build-css build-js",
     "eject": "react-scripts eject",
--- a/src/apps/AppManager.jsx
+++ b/src/apps/AppManager.jsx
@@ -6,7 +6,7 @@ import LaunchPage from '../components/LaunchPage';
 import LandingPage from '../components/LandingPage';
 
 export default class AppManager {
-    constructor() {  
+    constructor() {
 
         this.apps = [{
                 path: '/',
@@ -95,13 +95,13 @@ export default class AppManager {
             },
             {
                 path: '/ccp',
-                display: 'MCODE Pilot',
+                display: 'Compass',
                 app: CoreCancerPilotApp,
                 isExact: true,
                 dataSource: 'HardCodedReadOnlyDataSource',
                 patientId: '788dcbc3-ed18-470c-89ef-35ff91854c7g',
-                logoObject: { 
-                    path: './compass-logo.png', 
+                logoObject: {
+                    path: './compass-logo.png',
                     altText: 'Compass logo',
                     width: '45px',
                     height: '45px'
@@ -109,13 +109,13 @@ export default class AppManager {
             },
             {
                 path: '/ccp2',
-                display: 'MCODE Pilot',
+                display: 'Compass',
                 app: CoreCancerPilotApp,
                 isExact: true,
                 dataSource: 'HardCodedReadOnlyDataSource',
                 patientId: '788dcbc3-ed18-470c-89ef-35ff91854c7f',
-                logoObject: { 
-                    path: './compass-logo.png', 
+                logoObject: {
+                    path: './compass-logo.png',
                     altText: 'Compass logo',
                     width: '45px',
                     height: '45px'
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "watch-css": "npm run build-css && node-sass-chokidar src/ -o src/ --watch --recursive",
     "start-js": "react-scripts start",
     "check-version": "node ./scripts/check-version.js",
-    "start": "npm-run-all -p watch-css start-js check-version",
+    "start": "NODE_OPTIONS=--max-old-space-size=5120 npm-run-all -p watch-css start-js check-version",
```
dehall: This syntax doesn't work in Windows. One possible fix is to add the `cross-env` library and make the call `cross-env NODE_OPTIONS=--max-old-space-size=5120 npm-run-all -p watch-css start-js check-version`.  

That said, it looks like this is only necessary when using the larger mock data set (available separately per licensing), not the one that's in the repo.

noranda: @rdingwell Thoughts?

rdingwell: @dehall does the addition cause it not to run on windows?

rdingwell: @noranda the env setting crashes on windows.   Seeing how it's only needed large static data sets why dont we just remove it for now and set that env locally

dehall: Yes it prevents `yarn start` from running: 
> λ yarn start
yarn run v1.3.2
$ NODE_OPTIONS=--max-old-space-size=5120 npm-run-all -p watch-css start-js check-version
'NODE_OPTIONS' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

rdingwell: I'll go ahead an remove this and push the update

noranda: I removed and pushed.

dehall:
Apologies I didn't have time to review this in detail before having to jet off. Overall it looks good, but I saw one small issue.  When there's nothing in the "include only" section then the rows in the "compare with" section have inconsistent behavior.  Sometimes rows in the "compare with" section were just white bars all the way across, sometimes they have green sections that are unlabeled. 

screen shot 2019-02-11 at 8 28 38 am

Dtphelan1:
[DP1] Any reason we aren't using the typical react mechanism for event listeners like the `onClick` prop and are instead manually adding the event listeners ourselves on componentDidMount? [The motif I'm referring to]( https://reactjs.org/docs/handling-events.html )

Dtphelan1:
[DP2] There's 200px of bottom-margin on the `treatment-options-outcomes` class. Any reason why?

Dtphelan1:
[DP3] Is there any reason why the default "Similar Patient" options are too restrictive to display any comparative treatment options?

noranda:
@dehall Good catch! I pushed a commit that should fix this.

@Dtphelan1 
[DP1]: The reason it's done on component mount is because the user needs to click anywhere in the body to close the poppers, meaning we can't hook an onClick to a specific element.
[DP2]: I added this margin to prevent the screen from jumping when you open the compared table popper when you only have a few rows in there.
[DP3]: The seer data we're using in the repo is a small dataset, so it's very restrictive since we've now hooked up many of the checkboxes. It's useful for testing purposes, but not representative of the large datasets we expect to use. @rdingwell and I have a larger locally stored dataset that we use for demo purposes.

Rebased.

Dtphelan1:
[DP1] Makes complete sense -- my fault for not recognizing that click-anywhere-to-close behavior. Resolved!

[DP2] I see what behavior you're talking about when the popover overflows off the page. That said, the 200px of whitespace is pretty glaring both in the TDP and in the minimap. I don't think this should hold up your task, but I'd like to make a JIRA exploring writing a function that checks to see if the render is going to overflow off the page and maybe adjusts the `top` value to be a negative value, so that way the popover never causes that jumpiness.  That is, dynamically render something like this: 
image

[DP3] If you both have better demo data that you use locally then I feel okay, but it would be more convenient for others when testing/demoing if the default values we use now were not as restrictive, at least until we have that more expansive dataset deployed for everyone to use. I'll defer to @rdingwell or @gregquinn2001 here.

Dtphelan1:
Other than these comments, this all looks good to me. I'll approve once DP3 is addressed by Rob or Greg.

noranda:
[DP2] Agreed the extra space is a bit noticeable, although I probably wouldn't say glaring. I would caution against your suggested fix though, as the placement of the poppers helps the user to understand which table they're editing options for. If we move the compared options up, the user may be confused as to which options they're choosing (included or compared) even with the helper text. We should probably get some ideas from Dan, but agreed that this shouldn't hold up the PR.

[DP3] @rdingwell does have a larger dataset in his transfer folder, but we don't want to add it to the repo due to its size and licensing concerns.

Dtphelan1:
Re: DP3 I figured the data you mentioned wouldn't be used, the long term solution is querying a service. But until then, I defer to Rob and Greg on whether or not we should make the defaults less strict.   
  

  
Re: DP2, adjectives and connotations aside, it's a detail of the app's display that is a departure from the Mockups. Good point about the visual cues regarding the placement of the menus. Love the idea of chatting with Dan about potential alternatives. I'll make a JIRA tomorrow to implement whatever designs come out of that eventual discussion. Could you set up that meeting/communicate the nature of the problem to him? Thanks!
  

  
  

  
  
>

noranda:
[DP2] Emailed Dan for design advice, who reminded me that we will have a legend at the bottom which should solve this issue. That task is being worked on now.

Dtphelan1:
Sounds good to me!
  
  
  

  
  

  
  
>

noranda:
Rebased. Need a second approval on this. @dehall or @Dtphelan1 ?

Dtphelan1:
I'd be happy to approve once we address DP3! Pinging @rdingwell and @gregquinn2001 again for thoughts.

gregquinn2001:
I think it makes sense to create a JIRA to improve the data such that the initial default settings shows results. So for DP3, I think a separate JIRA makes sense. Thoughts?

Dtphelan1:
That would be a JIRA for DP3 then. Sounds good to me @gregquinn2001. I'll do that now and then approve.

Dtphelan1:
[DP-4] (I'm making a JIRA so not blocking but want to post here so we're all aware.) The rows for outcomes when viewing the portrait-mode ipad viewer on chrome look very dense and are (IMO) hard to read. We should investigate the responsiveness of this application in those cases. Image below: 

image

rdingwell:
Apologies for not seeing this sooner. I think Greg is right with regards to improving the default data for testing but I think making the initial filters less restrictive is fine too.  I’m fine with either solution

PR #573, opened 2019-02-18
Medication graph revamp

sgober:
Updated the medications visualizer based on the mockups in the JIRA (see ASCODCP-1674).  The `demo2` has the most medications and a good range of displays to look at for a first pass.  To test all the different variations, you can modify the inputs to the `RangeChart` component in `MedicationRangeChartVisualizer.jsx` instead of updating the patient.

NOTE: there is a pending decision about how to color the dot if there is no range defined and the value is not equal to the typical value.  @gregquinn2001 and I decided to make it red in this case pending the feedback from the clinicians. 

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/summary/RangeChart.jsx
+++ b/src/summary/RangeChart.jsx
@@ -52,45 +52,31 @@ class RangeChart extends Component {
             upperValueXPixels = ((this.props.upperValue - lowerBound) * numberOfPixelsPerUnit) + lineStartXPixels;
         }
 
-        // TO DO: add dot SVG
-        // let dotColor = 'black';
+        let dotColor = '#AAA';
+        let svgForDataPoint = null;
+        let svgForDataPointBorder = null;
+        if (!Lang.isNull(this.props.value)) { // only plot a dot if value is not null
 
-        // let valueXPixels = null;
-        // const distanceFromLineInPixels = 12;
+            let valueXPixels =  ((this.props.value - lowerBound) * numberOfPixelsPerUnit) + lineStartXPixels;
 
-        // let svgForDataPoint;
-        // // if value is null, don't plot a dot for it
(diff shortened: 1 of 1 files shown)


jafeltra:
This looks really good! I tried a bunch of combinations of values/ranges/typical values and everything looked really good. I also went through the mockups and checked all the cases outlined there and they looked good. All the tests pass. Two really minor things I noticed:

[JA1] This isn't necessarily something that needs to be changed because I don't think it's a real use case, but I noticed it and figured I'd at least mention it. If I have the lower value =  7, upper value =  7.5, typical value = 4, and value = 3, the typical value number doesn’t get printed on the chart. I think that the case it hits in the logic is supposed to omit the typical value number if there isn't room, but there is room on the graph. However, the typical value isn't within the range, so I don't know that this is a real use case, and it may be okay to keep the logic as is.

[JA2] I thought it was a little difficult to see the typical value tick mark when the value dot is on top of it. Is it possible to make it a bit taller? Although, looking at the mockups, it's not that pronounced there either, so it may be fine to keep as is.

sgober:
[JA1] This assumes that the typical value is between or equal to the upper or lower values.  It doesn't make sense for the typical value to be outside that range.  If I'm wrong on this then I can certainly change it. 

[JA2] I can certainly make it taller but I went off what the mockups said.  We can see if there's any more input on what's best.

jafeltra:
For JA2, you're right and I missed the dimensions on the left side of the mock ups. I'm fine with leaving both these issues as is! Nice work!

PR #574, opened 2019-02-18
Ccp table legend

KeeyanGhoreshi:
These changes add a legend to the ccp outcomes table.  The legend describes the little bar components that show the % of survivors for each treatment.  The topmost legend entry will change depending on which treatment is in context (being compared to).  

This also changes the way the bar graphs show negative change in survival.  Instead of a solid block of red that represents a decrease in survival, there is now a mask that is overlaid on top of the gray bar to show that it is a decrease in survival due to _comparison_.  This was causing issues with the legend, since the meaning of the gray part of the bar changed depending on whether there was an increase or decrease in survival.  If it increased, the gray bit represents the survival due to the treatment we are comparing to, and the green bit represents the increase due to the treatment we want to compare.  If it decreased, the gray bit would now represent survival due to the treatment we want to compare.  

In any case, the new red mask solves this problem.  There was a small bug in Chrome due to this change where the gradient used to make the mask would hang off the side of its background.  Simply put, the gradient was constructed using red "stripes" in the foreground and a background that matched the preceding gray bit to make it look like one cohesive segment.  

However, due to the way Chrome renders borders, the foreground was sometimes being pushed a pixel off the background and looking odd.  This was fixed by removing the border on the bar and adding a border to only the white bit of the bar. 

I'm a bit unsure what to name the red block on the legend though.  Right now it is "decrease in survival due to treatment" which isn't really true, since it makes it sound like the treatment is killing people.  I'm considering changing it to "comparative decrease in survival" but that's also not really a good description, especially considering how the other legend entries are phrased.  

There also appears to be a few merge conflicts due to this branch being made off of Noranda's branch and pulling along some of its commits from a while ago.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [ ] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
+++ b/src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
@@ -6,6 +6,7 @@ import _ from 'lodash';
 import { isSame, getCombinations } from '../../utils/arrayCompare';
 
 import BarChart from '../../visualizations/BarChart/BarChart';
+import TableLegend from '../../visualizations/TableLegend/TableLegend';
 import TreatmentsPopover from '../TreatmentsPopover/TreatmentsPopover';
 
 import './TreatmentOptionsOutcomes.css';
@@ -299,6 +300,7 @@ export default class TreatmentOptionsOutcomes extends Component {
                         )}
                     </div>
                 </div>
+                <TableLegend includedRow={includedRow} treatmentNames ={TREATMENT_NAMES}/>
             </div>
         );
     }
--- a/src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
+++ b/src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
@@ -300,7 +300,7 @@ export default class TreatmentOptionsOutcomes extends Component {
                         )}
                     </div>
                 </div>
-                <TableLegend includedRow={includedRow} treatmentNames ={TREATMENT_NAMES}/>
+                <TableLegend includedRow={includedRow} treatmentNames={TREATMENT_NAMES}/>
             </div>
         );
     }
--- a/src/mcode-pilot/visualizations/BarChart/BarChart.jsx
+++ b/src/mcode-pilot/visualizations/BarChart/BarChart.jsx
@@ -10,7 +10,18 @@ export default class BarChart extends Component {
         if (percentChange < 0) {
             // style for when survival decreases (red)
             mainStyle = {"width": `${survivedPercent}%`, "backgroundColor": "#9e9e9e" };
-            changeStyle = {"width": `${Math.abs(percentChange)}%`, "backgroundColor": "#e13949" };
+            changeStyle = {"width": `${Math.abs(roundedPercent)}%`, "background": `repeating-linear-gradient(
+                45deg,
+                transparent,
+                transparent 2px,
+                #EC9696 2px,
+                #EC9696 4px
+              ),
+              linear-gradient(
+                to bottom,
+                #9e9e9e,
+                #9e9e9e
+              )`  };
             textStyle = { "color": "#d9534f" };
         } else {
             // styles for when survival increases (green)
--- a/src/mcode-pilot/visualizations/BarChart/BarChart.jsx
+++ b/src/mcode-pilot/visualizations/BarChart/BarChart.jsx
@@ -5,7 +5,7 @@ import FontAwesome from 'react-fontawesome';
 import './BarChart.css';
 
 export default class BarChart extends Component {
-    getStyles = (survivedPercent, compareToPercent, percentChange) => {
+    getStyles = (survivedPercent, compareToPercent, percentChange, roundedPercent) => {
         let mainStyle, changeStyle, textStyle;
         if (percentChange < 0) {
             // style for when survival decreases (red)
@@ -44,7 +44,7 @@ export default class BarChart extends Component {
             compareToPercent = compareToNumerator / compareToDenominator * 100;
             roundedPercent = Math.floor(survivedPercent - compareToPercent);
             percentChange = survivedPercent - compareToPercent;
-            const styles = this.getStyles(survivedPercent, compareToPercent, percentChange);
+            const styles = this.getStyles(survivedPercent, compareToPercent, percentChange, roundedPercent);
             mainStyle = styles.mainStyle;
             changeStyle = styles.changeStyle;
             textStyle = styles.textStyle;
@@ -66,6 +66,7 @@ export default class BarChart extends Component {
                     <div className="progress-bar">
                         <div className="prog-fill" style={mainStyle}></div>
                         {hasCompare && <div className="prog-fill" style={changeStyle}></div>}
+                        <div className="prog-fill-empty"></div>
                     </div>
                 </div>
             </div>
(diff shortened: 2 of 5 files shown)


Dtphelan1:
[DP1] There are some console warnings that should be addressed. Namely an import of FontAwesome that doesn't get used and a space before an `=` that shouldn't be there.

Dtphelan1:
Open Question - It strikes me that we use the words 'additional survivors' for the Green-bar, talking about people, but use the words 'decrease in survival' for the Red-bar, referring to the percentage instead of the people. I think I'd like to harmonize the language here -- either talk about the survival rate, or the survivors. Do people feel strongly?

kghoreshi:
> Open Question - It strikes me that we use the words 'additional survivors' for the Green-bar, talking about people, but use the words 'decrease in survival' for the Red-bar, referring to the percentage instead of the people. I think I'd like to harmonize the language here -- either talk about the survival rate, or the survivors. Do people feel strongly?

I have no strong feelings, but talking about percentages seems to me to be more accurate, since the red and green bars represent percent increase, not an increase in individual survivors.

Dtphelan1:
[DP-X] Can make a JIRA for these if needed, but regaring the array compare file: 
    
    a) ArrayCompare is feeling more like an ArrayUtil file at this point
    b) getCombinations is indecipherable to me. I'd be happy to replace it with something more human readable but that function is pretty hard to read with all the bitwise operations going on.

Dtphelan1:
@KeeyanGhoreshi 's question regarding how to describe to the red-bar change, IMO, comes from an either a miscommunication or an inconsistency in what we're visualizing. The title of that subsection suggests we're showing the combed efficacy of the currently selected treatment (let's say "Surgery") with other selected treatments. But that can't _always_ be the case, right? For example, we can't combine "Surgery" with "none (actively monitoring)". 

That suggests to me that either 1) we aren't showing combined treatment plans in the first place, or 2) that we're showing two kinds of information: combined treatments when possible with the current treatment, visualizing the impact those additions would have, and treatments that _cannot_ be combined with our current treatment, juxtaposing their expected outcomes with the other selected treatments? I'm looking in the code now to confirm which it is, but thought I'd articulate my thoughts here. 

Back to Keeyan's original question: If we're displaying 1), a candidate for what the red-bar legend says could be "decrease in survival due to alt. treatment." If it's 2), some candidates for the red-bar legend could be general as Keeyan suggested ("Comparative decrease in survival"), or super specific ("Decrease in survival due to alternative treatment or due to combining treatments").

Another option might be to to avoid displaying both kinds of outcomes in the same subsection (one way -- not necessarily _the_ way -- would be to have a "combined with" subsection and an "alt treatment" subsection.) Spit-balling here. 

`tl;dr` red-bar labeling is hard to do, but that might be due to the fact that we're not always combining treatments and need to disambiguate between decrease in survival outcomes because of alternative treatments vs. combined treatments.

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
```diff
@@ -147,34 +183,107 @@ export default class TreatmentOptionsOutcomes extends Component {
...
+        );
+    }
+
     render() {
+        const { includedTreatments, comparedTreatments, includedOpen, comparedOpen } = this.state;
+        const { similarPatients } = this.props;
+        const includedTreatmentPatients = similarPatients.filter(patient => isSame(patient.treatments, includedTreatments));
+        const comparedTreatmentCombinations = getCombinations(comparedTreatments);
+        const includedRow = this.generateRow(includedTreatmentPatients, includedTreatments);
+        // TODO: move this to componentWillReceiveProps
```
Dtphelan1: [DP3] Should this be moved to ComponentWillReceiveProps?

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.scss
```diff
@@ -1,5 +1,7 @@
 $color-gray-light: #bbb;
 $color-gray-lighter: #eee;
+$color-blue: #0084b8;
```
Dtphelan1: [DP4] Could we use the values for colors defined in the variables.scss file?

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.scss
```diff
@@ -1,5 +1,7 @@
...
+$color-blue-light: #ecf3f5;
```
Dtphelan1: Probably a separate JIRA, but we should follow up with Dan Reeves to see how/if this fits in with our overarching style-guide. If it does, we can add it to our variables file accordingly and reference from there.

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
```diff
@@ -3,46 +3,101 @@ import PropTypes from 'prop-types';
...
+import { isSame, getCombinations } from '../../utils/arrayCompare';
+
 import BarChart from '../../visualizations/BarChart/BarChart';
+import TableLegend from '../../visualizations/TableLegend/TableLegend';
+import TreatmentsPopover from '../TreatmentsPopover/TreatmentsPopover';
 
 import './TreatmentOptionsOutcomes.css';
 
-const NO_TREATMENT = 'none (actively monitoring)';
+const TREATMENT_NAMES = {
```
gregquinn2001: [GQ1] In TreatmentOptionsOutcomes, why are we directly listing treatment options and side effect names specific to our current data? The list of treatment options and side effects should be ascertained from the data returned from the treatment options service - not hard coded! We had this in the treatment options viz we did originally.

KeeyanGhoreshi:
[GQ1], [DP1], [DP2], [DP3], and [DP4] are all referencing changes made in @noranda's commits, which have already been merged to master in #568.  She could speak better to these issues than I can.

gregquinn2001:
GQ1 at least was showing up in the changes to files for this pull request???

KeeyanGhoreshi:
The only commits that should really be in this PR are the two that I've committed, "added a legend" and "fixed small gradient related bug".  This is probably an issue with the way I made this branch or opened this pull request, sorry.

Dtphelan1:
Hard to tell right now how many of the DP's are still relevant since this branch has yet to be rebased onto master. I think some have been addressed in future commits on noranda's branch, but it's hard to tell in this state. We can determine which of the DP's need to be made into JIRA's once that happens.

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.scss
Dtphelan1: Created a JIRA

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.scss
Dtphelan1: Talked with Greg and made a JIRA

Dtphelan1:
[DP1] - I fixed this during the rebase
[DP2] - Made a JIRA
[DP3] - Was actually fixed on Noranda's branch earlier; OBE after rebasing
[DP4] - Created a JIRA for this and the styleGuide unification
[GQ1] - Created a JIRA

Still want to hear back on my meta-level comments about the designs, and whether or not we are _just_ showing combined treatments or something else, as I think this might contribute to the confusion re: the red bars.

nicoleng12:
Everything looks good on my end. Backend tests pass, this works in both IE and chrome, and Dylan covered the issues in his comment above.

PR #578, opened 2019-02-25
Integration of mCODE v0.5 ES6 Classes

danlee1025:

Addresses JIRA-1684.

- Imported mCODE v0.5 ES6 classes to be used as Flux Notes Object Model.
- Created `HardCodedMcodeV01DataSource` to load mCODE v0.1 JSON and map to Flux Notes Object Model.
    -  `EntryMapper.js` exports `mapEntries` function that performs mapping and is used by the datasource.
-  Updated wrappers and UI to use new ES6 classes.

Part of this task was to make future updates to mCODE easier to incorporate so we have decided to use mCODE v0.5 as our base model for the Flux Notes Object Model.  For future updates of mCODE, we will only have to write the mapper to map the updated mCODE entries to the Flux Notes Object Model, which should minimize the amount of changes we need to make to the UI and wrappers.

While testing, make sure to test all parts of the app as these changes affect everything and bugs can be subtle.  I know that people have run through the pre-release script for all the endpoints as subtasks but changes have been made since then.

Big thanks to everyone who helped find/fix bugs on this branch!

## Submitter:

**Running the Application**

**Task**

**Documentation**

- [x] Pre release script is updated on the [Pre Release Testing Script wiki page](https://github.com/FluxNotes/flux/wiki/Pre-Release-Testing-Script) (add/remove single features to test, do not run through full script)
- [x] Update the list of key features with new features on the [Key Application Features wiki page](https://github.com/FluxNotes/flux/wiki/Key-Application-Features)
- [x] Document any new APIs/extension points

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


Dtphelan1:
image

Dtphelan1:
[DP-1] New names for patient files have been lost.

jafeltra:
[JA1] #ER is broken. I get the following error:
Uncaught TypeError: Cannot read property 'apply' of undefined
    at SingleHashtagKeyword.callMethod (SingleHashtagKeyword.jsx:210)
    at SingleHashtagKeyword.jsx:278
    at Array.forEach (<anonymous>)
    at SingleHashtagKeyword.updatePatient (SingleHashtagKeyword.jsx:277)
    at SingleHashtagKeyword.FullApp._this.handleShortcutUpdate [as onUpdate] (FullApp.jsx:299)
    at SingleHashtagKeyword.setAttributeValue (SingleHashtagKeyword.jsx:162)
    at EntryShortcut.jsx:61
    at Array.forEach (<anonymous>)
    at SingleHashtagKeyword.initialize (EntryShortcut.jsx:58)
    at Object.FullApp._this.newCurrentShortcut (FullApp.jsx:289)

jafeltra:
I also noticed that @referred by and @recent imaging are broken on /demo2. They're also broken on master too though, so don't need to be fixed now.

dehall:
[DH1] looks like the ES6 classes were generated against FHIR STU3 instead of DSTU2. For an example, in DSTU2 the Condition resource has a `patient` and an `encounter` but in STU3 these get renamed to `subject` and `context` . Here the classes that map to FHIR Condition, such as shr.base.ConditionPresentAssertion have fromFHIR methods that look for `subject` and `context` so we know it's based on STU3) For the immediate future we want DSTU2, so the classes will need to be regenerated. Hopefully that's a simple drop-in replacement since the object model won't change, just the fromFHIR methods. Also the Coding/CodeableConcept classes have to be manually modified: https://github.com/FluxNotes/flux/wiki/Updating-Generated-ES6-classes

src/dataaccess/mcodev0.1-datasource/model/procedure/FluxProcedurePerformed.js
```diff
@@ -0,0 +1,46 @@
+import TimePeriod from '../shr/core/TimePeriod';
+import ProcedurePerformed from '../shr/procedure/ProcedurePerformed';
+
+class FluxProcedurePerformed {
+    constructor(json) {
+        this._procedurePerformed = ProcedurePerformed.fromJSON(json);
+    }
+
```
dehall: This class (and the other `FluxProcedurePerformed` class in the v0.1 class structure) needs an `entryInfo` getter that gets the entryInfo from the contained `_procedurePerformed` , otherwise it can crash the reference lookup in PatientRecord.getEntryFromReference. At a glance I don't see any others where it's missing but someone please double check me on that.

danlee1025: Thanks, I pushed up the getters.  I don't think we have any instances of `ProcedurePerformed` so didn't catch this in our endpoints.

danlee1025:
JA1 should be resolved.

src/dataaccess/mcodev0.1-datasource/EntryMapper.js
```diff
@@ -0,0 +1,900 @@
...
+
+const mapCancerDisorder = (entry, klass) => {
+    const newCondition = new klass();
+
+    mapEntryInfo(entry.entryInfo, newCondition);
+    newCondition.patient = mapReference(entry._condition.patient);
+    newCondition.anatomicalLocation = entry._condition.anatomicalLocation.map(a => mapAnatomicalLocation(a));
+    newCondition.category = mapPassThrough(entry._condition.category, Category);
+    newCondition.clinicalStatus = mapClinicalStatus(entry._condition.clinicalStatus);
+    newCondition.findingResult = new FindingResult();
```
dehall: I think this might not be the right field to use here.
`ExistenceAssertion` (and so its child classes `ConditionPresentAssertion`, `CancerDisorderPresent`, etc) has a field `findingResult` which should be a `shr.base.PresentOrAbsent` (restricted to either "Present (qualifier value)" or "Absent (qualifier value)"), and a `findingTopicCode` which should be a `shr.base.ConditionOrDiagnosisCode`
It looks like the fixtures and UI use the `findingResult` field when they should use `findingTopicCode` for the actual code of the condition which is in `condition.value` in the old model.

danlee1025: Talked to Dylan and pushed up a fix for this.

Dtphelan1:
Question for @daniellee1025: Should FluxTNMClinicalPrimaryTumorClassification extend Entry in the FluxTNMClinicalPrimaryTumorClassification.js file?

danlee1025:
@Dtphelan1 It can if you need.  I usually don't extend `FluxEntry` unless I want to inherit the `sourceClinicalNoteReference` getter but maybe we should have all our entry wrappers extend it?

Dtphelan1:
I'm not sure -- either way, I think there are a couple of strategic decisions from the modeling perspective that I don't know where to find documentation for/don't know if there is any documentation for at all. It might be valuable to get that knowledge on paper at some point. We can chat more about that tomorrow though 👍

gregquinn2001:
Every wrapper that represents an entry should extend FluxEntry ultimately, but we don't always because as Dan said, the only functionality you get is the source clinical note.

gregquinn2001:
No reason for Dan to worry about making every wrapper of an entry extend FluxEntry. If we decide we need that, it would be a separate task.

Dtphelan1:
Sounds good to me.

src/apps/AppManager.jsx
```diff
@@ -98,7 +98,7 @@ export default class AppManager {
                 display: 'Compass',
                 app: CoreCancerPilotApp,
                 isExact: true,
-                dataSource: 'HardCodedReadOnlyDataSource',
```
dehall: If the HardCodedReadOnlyDataSource is replaced everywhere, is there any reason not to just delete it altogether? It references the same patient jsons as the McodeV01DataSource, but doesn't wrap them in EntryMapper.mapEntries so if it does still get used somewhere it probably isn't going to work.  I notice that it is still referenced as a fallback in case the datasource is not defined in props in FullApp:
```
        if (Lang.isUndefined(this.props.dataSource)) {
            this.dataAccess = new DataAccess("HardCodedReadOnlyDataSource");
        } else {
            this.dataAccess = new DataAccess(this.props.dataSource);
        }
```

danlee1025: Good catch, I'll delete this.

PR #586, opened 2019-03-01
SMART on FHIR Data Source

dehall:
This PR covers the following JIRA tasks:
- 1512: Create SMART FHIR Data Source
- 1603: Update Colorectal FHIR patient and populate FHIR server 
- 1698:  Test the SMART on FHIR DataSource with Colorectal FHIR patient
- 1788: Make Patient Control Panel handle missing data from SMART-on-FHIR data source

Adds a SMART on FHIR data source to be used by the SmartApp, so that patients can be loaded from a SMART on FHIR server. The information flow is a little different from usual:
1. The user starts in their EHR
2. The user chooses to launch a SMART on FHIR app. (in this case FluxNotes/Compass)
3. The app redirects to the SMART endpoint to authenticate/authorize access to the patient data
4. The SMART endpoint redirects back to the app with a token
5. The app uses that authorization token to fetch patient data from the server

As of this implementation, the data source itself and the fromFHIR methods it leverages require that the incoming FHIR be profiled with mCODE profile URLs. Any FHIR resources that don't have mCODE profiles will be dropped. So there are 2 ways this is accomplished:

#### 1. Records can have mCODE profiles already
The easy way is when the record is already profiled with mCODE profiles. I've loaded some test patients into an internal instance of the SMART on FHIR sandbox, all with the last name "McOde123"

To test this approach, start from the internal SMART launcher (should set everything correctly but in case not, select Provider EHR Launch, FHIR Version R2, and app launch URL `http://localhost:3000/launch` or `http://<wherever your local instance is>/launch`  )
http://moonshot-dev.mitre.org:4001/?auth_error=&fhir_version_1=r2&fhir_version_2=r2&iss=&launch_ehr=1&launch_url=http%3A%2F%2Flocalhost%3A3000%2Flaunch&patient=&prov_skip_auth=1&provider=&pt_skip_auth=1&public_key=&sb=&sde=&sim_ehr=0&token_lifetime=15&user_pt=

then click Launch App, login info not necessary so just click Login, then choose any of the McOde family from the patient selection. It should redirect to your local app and show patient info like birthdate, place, and conditions.

#### 2. The shim adds mCODE profiles per mapping rules
The challenge is that records out in the wild are not already profiled with mCODE profiles. That's where the shim comes in: https://github.com/standardhealth/proxy-shim
This PR includes a config setting that will reroute calls to specified servers to a shim instead.

In `AppManager.jsx`:
```js
            {
                path: '/launch',
                display: 'Flux',
                app: LaunchPage,
                isExact: true,
                launchContext: {
                    client: {
                        client_id: '6c12dff4-24e7-4475-a742-b08972c4ea27',
                        scope:  'patient/*.read user/*.* openid profile',
                        redirect_uri: 'http://localhost:3000/smart'
                    },
                    // uncomment for shim
                    // server: "http://localhost:3001/1_0_2"
                }
            },
```

The inner workings of the shim are out of scope for this PR, so one way to test it is to configure FN to use the shim and then just ensure that connections are made to the shim and not to the original FHIR server.
Alternatively I will work with reviewers to setup a working shim so that it will add the appropriate profiles.

In this approach, start from some other SMART launcher (recommend using the public SMART on FHIR sandbox at http://launch.smarthealthit.org/?auth_error=&fhir_version_1=r2&fhir_version_2=r2&iss=&launch_ehr=1&launch_url=http%3A%2F%2Flocalhost%3A3000%2Flaunch&patient=&prov_skip_auth=1&provider=&pt_skip_auth=1&public_key=&sb=&sde=&sim_ehr=0&token_lifetime=15&user_pt= , same guidance as above)

In this instance there are no McOdes to choose so just pick any patient. Then watch network activity to ensure that the calls to the FHIR server go to the shim and not the original server associated with the launcher. If the shim is setup to add mCODE profiles, FN should display the patient as though it did.

Notes:
 - This only adds loading of patients, not saving. Saving would require the toFHIR methods to be in place, but they were very broken so they were removed from the ES6 class generator. That said I think I know how it would work so I added that code in a comment where it should go.
 - This adds new dependencies: `nock`, used for mocking network requests and `mock-local-storage` which lets the FHIR client library use sessionStorage in tests
 - Not everything in the UI currently plays nicely with patients loaded from FHIR. ~The most obvious example is that patient names are missing. This will be fixed among other things in JIRA task 1788 (or I can include it here if people want me to)~
 - This includes fixes and workarounds for some open issues with the ES6 class generator: https://github.com/standardhealth/shr-es6-export/issues
 - The change to SmartOnFhirLaunchContext.json is required to make this work on IE. The relative URL was fine on Firefox/Chrome but IE didn't like it.
 - Apologies for any unnecessary whitespace changes. I had my editor set to remove whitespace at the end of lines for a while until it caused too many diffs, but I guess I didn't undo all of them.
 
---

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


gregquinn2001:
[GQ1] I chose Mary Jane35 McOde123 and did not get name showing up in patient banner

gregquinn2001:
Overall though it's awesome to see patients coming from SMART on FHIR sandbox being displayed!

gregquinn2001:
[GQ2] We are getting an active condition for numerous McOdes that has missing data showing for its condition name. Can we fix it or create a JIRA to fix it if we think it's out of scope?

gregquinn2001:
[GQ4] The apps and data sources should not know anything about the Shim. It is an mCODE FHIR service that the SMART on FHIR data source can be configured to talk to. The instance of the shim should be configured to talk to a specific EHR IMO. To that end, the data source you created should be called SMART on FHIR mCODE data source. Maybe even have a version in there. We can create another data source that knows how to talk to a generic SMART on FHIR implementation and can do declared mapping of the expected FHIR resources (based on particular EHR instance being talked to) to mCODE FHIR. That should be a different data source IMO! Let's keep the ideal one pure and simple - the SMART on FHIR mCODE one.

public/ServerConfig.json
```diff
@@ -1,3 +1,15 @@
 {
-    "baseURL": "http://localhost:3001/api"
+    "baseURL": "http://localhost:3001/api",
+    "fhir" : {
+        "shimServerOverrides": {
+            "http://example.com/fhir": "http://localhost:3001/3_0_1"
+        },
+        "supportedResourceTypes": [
```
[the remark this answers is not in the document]
rdingwell: That should be doable from the shim side as well seeing how it generates a conformance statement based on the resource endpoints that have been added.

dehall:
Re GQ4: Let's sync up offline with @rdingwell but my understanding is there are 2 reasons for the shim: 1) applying profiles or otherwise modifying FHIR resources between the EHR FHIR server and SoF app, and 2) integrating data from other non-FHIR sources. Creating a separate data source that can handle non-mCODE-profiled resources addresses 1 but not 2, so as far as I know the concept of a shim is still needed. So then we can rename and rework things a little bit, but based on how the SMART-on-FHIR process works, knowledge of the shim has to exist here in the app. The EHR launches the app and tells it "get FHIR data at this endpoint" and has no knowledge of the shim. So the app has to be configured somehow to use the shim URL instead of the original URL.

dehall:
Re GQ2: I suspect the ones with missing data are the cancer disorders. I'll look into what's going on though. If the issue is the sample patients are bad then I'll fix them right away, or if it's a quick display fix I can make that here too. If it looks like it involves a lot of rework to existing code then I'll check back to see what people want to do.

public/ServerConfig.json
```diff
@@ -1,3 +1,15 @@
 {
-    "baseURL": "http://localhost:3001/api"
+    "baseURL": "http://localhost:3001/api",
+    "fhir" : {
+        "shimServerOverrides": {
```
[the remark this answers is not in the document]
rdingwell: The javascript smart client is a global object and does not act like an instance but rather a singleton.    There may be a way to remove the use of the config file and move the information into the meta data contained in the app manager but it will require some reworking.  

This can be overridden in parameters passed into the window.FHIR.oauth2.authorize method being called in the flux Launch component .  The shim server already supports returning the oauth2 urls required for authenticating to the server so we would need to redirect the client to use the shim server as a baseurl in the Launch component.    Mapping the iss params passed in to a specific shim server depending on the Flux app(s) configured to use Smart on FHIR will require either  multiple Launch endpoints with the override provided in the metadata in AppManager or a single Launch endpoint that uses a config file such as ServerConfig that contains the mappings.   This may require similar mappings for the flux SmartApp as well.  

I have no real preference on either way just as long as it is testable.  We can sync up and decide on which is the best approach to take.

dehall:
GQ1 and GQ2 should be addressed as of the latest commit. Looking into what's possible for GQ3 now.

public/ServerConfig.json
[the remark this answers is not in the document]
gregquinn2001: Sorry. I'm not quite sure I followed everything in your comments so a discussion probably makes sense. @dehall can you set up a time for the 3 of us to talk? Actually, let's make it 4 and include Dan. Definitely add a skype if it's today! :) Basically, I think the shim should be a SMART on FHIR server and that Flux Notes should not be aware of whether it's talking to the shim or some EHR that supports mCODE.

gregquinn2001:
@dehall I'm not sure we're communicating on GQ4. I understand the Shim is meant to be capable of serving up mCODE FHIR regardless of how the EHR provides the data to the Shim. My point was that the client side code should not be aware of the Shim. The data source will have a configured URL (for mCODE FHIR) which we will provide to authorize and then use for subsequent FHIR calls through the FHIR client. There should be no need for the data source to be aware of the shim at all - just have the ability to override the iss via app manager configuration. So I think our solution as discussed on the phone addresses GQ4 the best we can.

dehall:
Yes my last comment was from way earlier this morning, before our call. As of now it should reflect the approach we discussed, the last remaining thing is I'm trying to determine the best way of testing it. The old approach was fairly self-contained, now it spans multiple components and has more external dependencies.

dehall:
@gregquinn2001 Addressed GQ1-4 and rebased off master. I still have a couple test cases I'm still trying to get working locally but this should be good for re-review.

gregquinn2001:
[GQ5] Dates in targeted data panel (e.g., diagnosis dates) now are showing times. I thought we had formatting on those that only showed the date...

gregquinn2001:
Do we support SHRO profiles? We don't just want mCODE/Oncocore data elements. I'm assuming yes because we see some demographics

gregquinn2001:
Also, do we support all of mCODE? If we had one of test patients in the sandbox with mCODE/SHRO profiles, what data would be not see right now that we see in the normal Flux Notes?

gregquinn2001:
[GQ6] Also Dylan, can we name the data source consistently with others (e.g., HardCodedMcodeV01DataSource) so it should be named McodeV05SmartOnFhirDataSource.

dehall:
Addressed GQ6 in latest commit.

Re GQ5: It looks like all the dates in all the current test patients are stored as "DD MM YYYY" and the UI is just displaying them as-is. In the case of Condition diagnosis dates, the field is `ConditionPresentAssertion.onset` which is a `shr.base.Onset` where the value could be any of the following types: dateTime, shr.core.Age, shr.core.TimePeriod, shr.core.Range, or string. ( https://github.com/standardhealth/shr_spec/blob/6d5746f87e9a4976bf2d3b5c8bd22e72196014e5/spec/shr_base_finding.txt#L356 )  I can either make a change here to handle those different types and consolidate them into a single format, or create a jira for that.

gregquinn2001:
For GQ5, the diagnosis date (onset) should not have a time component so the data source should drop the time component if there's one coming in as dateTime when it stores the value in the flux notes object model. What do you think?

dehall:
Should the data source be destructive like that? Since we eventually want this to be able to save back to the FHIR server, I would think we'd want to preserve everything as-is and only display the parts we care about. Maybe we should sync up offline

gregquinn2001:
Good question. I think the data coming in is unrealistic. An EHR is not going to assign a diagnosis time. The creation time of the record is a date and time, but the diagnosis date should not be misleadingly portrayed as that accurate. If it has a time coming in and we do an update then the time will get dropped. I'm ok with that, but we can talk to May Terry about it if you want.

rdingwell:
Is there a need to keep the mcodev0.1-datasource around?

src/model/json-helper.js
```diff
@@ -58,9 +58,17 @@ export function getNamespaceAndName(json={}, type) {
...
 export function getNamespaceAndNameFromFHIR(fhir, type) {
   // Special case for primitives
-  if (typeof fhir !== 'object' && type != null && type.indexOf('.') === -1) {
-    return { namespace: 'primitive', elementName: type };
+  if (typeof fhir !== 'object') {
+    if (type == null) {
+      return { namespace: 'primitive', elementName: typeof fhir };
+    } else if (type.indexOf('.') === -1) {
+      return { namespace: 'primitive', elementName: type };
+    }
```
rdingwell: I'm going to guess that this is a file that comes from the SHR toolchain but it seems like there is a condition that is missed here.  If typeof fhir !== 'object' and type.indexOf('.') !== -1 then nothing is returned.  This is probably a case where you would not expect the type to be some kind of object hierarchy so maybe an error should be thrown to prevent null pointers upstream.   

If this is an shr toolchain file then we can comment on that there.

rdingwell:
If the dates are unrealistic then we should be able to just adjust the test data.  If real systems do not supply the time components than everything should be fine,  if they do provide time components I don't think that they should be dropped.

dehall:
Discussed open questions with Greg. For posterity:

> Do we support SHRO profiles?

Yes, this data source supports all profiles in the SHRO model, not just oncocore. For instance all non-cancer conditions are shr.base.ConditionPresentAssertion. 

> Also, do we support all of mCODE/SHRO? 

Yes, the data source supports the entire model. One thing we don't see right now is procedures in the targeted data section, but that's because the section looks for ProcedureRequested and all the test patients have ProcedurePerformed. Jira item ASCODCP-1820 will address that.

> [GQ5] Dates in targeted data panel (e.g., diagnosis dates) now are showing times. 

This will be addressed in 2 new Jiras:
 - one to add a date formatter to the diagnosis date getter
 - one to handle the various types of objects that the Onset value could have, in addition to strings


Action Item: [DH1] I need to review the McOde patients to make sure they all have a cancer condition as expected

src/model/json-helper.js
dehall: In the case where typeof fhir !== 'object' and type.indexOf('.') !== -1 then it will fall through to the logic below and eventually throw an error. I can make that more explicit with a separate branch or comment if you'd like.

This file does come from upstream in the shr toolchain and these changes should eventually make their way back up, but it was quicker to put them in place here first.

src/dataaccess/McodeV05SmartOnFhirDataSource.jsx
```diff
@@ -0,0 +1,124 @@
...
+                } catch (e) {
+                    // just log the error, don't stop processing other potentially good objects
+                    console.error(e);
+                    return null;
+                }
+            });
+
+            allObjects.push(...referencesOut);
+            const json = allObjects.filter(o => o).map(o => o.toJSON());
+            return new PatientRecord(json);
```
rdingwell: The way this looks to work is to use ObjectFactory to create the mcode objects,  these are then serialized to json, passed off to the PatientRecord which then calls the FluxObjectFactory which also delegates to the ObjectFactory to turn them back into Objects.   Is there a reason for the creation -> serialization -> creation   workflow?

dehall: It's because the PatientRecord contains Flux wrappers (ex FluxPatient, FluxPerson, etc) not just the plain ES6 model classes, and the wrappers don't contain support to load them from FHIR. It would be nice to be able to support FHIR -> Flux wrappers directly, but that would require modifying all of the wrapper classes and object factories, then everything that calls them, etc.

src/model/json-helper.js
rdingwell: Ok, I had some odd indentation issues where it looked like the method ended before it actually did.

dehall:
> Is there a need to keep the mcodev0.1-datasource around?

This is currently being touched in the "Align Cancer-Summary Section with mCODE Data" PR, but as of now this is still needed because the demo patients are still in v0.1 format, so those data models are used to load those objects, and then they get converted to v0.5 via the EntryMapper. So short answer yes they are still necessary.

dehall:
Turns out the SMART sandbox FHIR server only returns 10 results at a time for queries, so for certain patients that have > 10 conditions the cancer condition doesn't get included. There's plenty of room for discussion on how much data the client should fetch and whether it should be configurable, but at a minimum here I want to figure out how to get all data, and then we can iterate from there if necessary.

dehall:
Addressed [DH1] aka the fact that not all patients had cancer conditions. Added logic that checks for when the server indicates that there are more search results and follows those links. The fhirclient library has a `fetchAll` function that does basically the same thing, but it has a bug that makes me not want to use it, so this is a local implementation. 

All open issues should now be resolved.

dehall:
Update: pushed one more change to not render the MRN `<span>` and empty parenthesess `()` if the record doesn't have a MRN. This PR now covers the entirety of JIRA 1788 "Make Patient Control Panel handle missing data from SMART-on-FHIR data source", so reviewers please take a look from that perspective as well. @rdingwell @gregquinn2001

gregquinn2001:
Did the McOde family get deleted from smart on fhir sandbox? I can't find them?

gregquinn2001:
@dehall  It's working again. The previous default dr is back and then I find McOdes in patient selection dialog.

gregquinn2001:
[GQ7] Sorry Dylan! I'm assuming the McOdes have medications but they aren't showing up. That may be because the FHIR data does not include any linkage between medication and condition?

gregquinn2001:
[GQ8] No procedures? Is that because there's no linkage between condition and procedures like Flux Notes expects?

dehall:
Re GQ8, the procedures aren't listed because the UI is looking for ProcedureRequested but all the FHIR Procedures get converted into ProcedurePerformed. I added Jira 1820 for the UI to support both ProcedureRequested and ProcedurePerformed.

Re GQ7, I noticed that too and I guess I got distracted and forgot. There are 2 reasons the medications aren't showing up: 1) I didn't add the right mCODE profile to the medication resources, and 2) the prescription date field that the UI looks for isn't used by synthea patients. I can't say which one is "right" here so my suggested fix is to just make the synthea data use the field the UI expects. I can fix both and update the patients on the server.

gregquinn2001:
Sorry Dylan. GQ8 was already discussed between us. My memory is awful!

For GQ7, let's figure out which is right based on what the SHR spec team says? Then let's fix whichever one is wrong?

dehall:
Re GQ7: there's even a third reason they aren't showing up, the MediationRequested.fromFHIR method doesn't handle the FHIR reasonReference field which is how the Medication and Condition are linked in FHIR. So I'm going to investigate that as well

rdingwell:
@dehall can you just patch what is there for the generated models wrt the reasonReference field?  I think any significant work on the exporter code may be a waste as the SHR core is made more in line to FHIR than CIMI.  

What happens to meds that have no reason reference associated with them?   Is that they are just not displayed?  Is the requirement that there be a reason associated with the med something that is definitely going to be reflected in real world systems from a FHIR point of view?

dehall:
@rdingwell Yes patching the generated model is the current plan. 

@gregquinn2001 and I discussed the fact that Medications without a reason reference aren't going to show up, and that's something we should revisit one we start to know what the real data is going to look like. Since a reason for the medication is required for billing purposes, our current thinking is that a reason "should" always be there. Should I make another Jira for a potential change there?

Also, during investigation of the reasonReference field I found another bug in the generated models around how they locate referenced objects. So I'll open another Jira for that and handle it specially in the reasonReference case.

dehall:
Here's the latest status.
Everything should be resolved except GQ7: Medications not showing up. This is a combination of several factors:
 1) The sample patients' MedicationOrder resources don't have the right profile on them
 2) MedicationOrder.reasonReference isn't mapped in fromFHIR, this seems to be a spec or other toolchain issue as it's not even included in the profile that gets passed to the ES6 exporter
 3) There are some issues with references and IDs in the fromFHIR logic
 4) The UI for Medications checks for prescription date in the field FluxMedicationRequested.expectedPerformanceTime, but the test patients don't have data that maps into this field. Synthea patients instead have a field MedicationOrder.dateWritten (which is required by the argonaut profile, so I would suggest the UI at least consider that as a fallback).

@gregquinn2001 I'm happy to address as many of these as you think makes sense here, though Medications aren't going to show up unless all of them are resolved or at least addressed with workarounds. So my suggestion is to make a couple more tweaks in support of issue 3 here, but defer the rest to a separate jira to address Medications. (and then rebase)

gregquinn2001:
@dehall  sounds good! Thanks for the complete summary! It helped me understand the issue better. Let's go with your suggestion!

gregquinn2001:
Let me know when you're done with your changes for issue 3 above and I'll approve.

dehall:
@gregquinn2001 pushed and rebased

PR #589, opened 2019-03-04
changed comparison to combination

KeeyanGhoreshi:
This PR changes the CCP table from comparing treatment options to combining them.  If you select surgery as your included treatment, instead of seeing `surgery & radiation therapy` in the table, you should see just `radiation therapy`.  Previously this would mean that `radiation therapy`, alone, was being compared with `surgery`, alone.  Now it shows the difference in survival between `surgery` and `surgery` combined with `radiation therapy`.  

There's also some changes to `arrayCompare` and `getCombinations`.  Despite the usage of bit shifting I found the current implementation of `getCombinations` to be  the most intuitive solution to power sets I could find.  Essentially, all its doing is counting in binary.  For example, if we have `[a, b, c]` and we want the powerset, we know powersets have size 2<sup>L</sup> where L is the length of the array. In this case the length is 3, so the function just does this:

| i  | binary | set |
| -- | ----|-------|
| 0 | 000 | `[]`|
| 1 | 001 | `[c]`|
| 2 | 010 | `[b]`|
| 3 | 011 | `[c, b]`|
| 4 | 100 | `[a]`|
| 5 | 101 | `[a, c]`|
| 6 | 110 | `[a, b]`|
| 7 | 111 | `[a, b, c]`|

I left this solution as is in the code and left some comments, but if we're still concerned about readability this same thing could be done with some `for` loops.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


Dtphelan1:
[DP1] - Looks like tests are failing

Dtphelan1:
Also, not trying to _intentionally_ undo all the hard work put into understanding the bitwise computation of powersets, but...
[DP2] - If we're strictly looking at combinations and not comparisons with all possible permutations of treatments, we don't need to look at the powerset of possible treatments anymore -- we just need to look at the set of combinations of each additionalTreatments with the selectedTreatment, something like: 
```
selectedTreatment = 'surgery'
allTreatmentCombos = []
for t in additionalTreatments: 
    allTreatmentCombos.push([t, selectedTreatment])
```

I'm sure there are more elegant solutions to that, but the gist is that we don't need to compute the whole powerset if we're just going to filter down to those treatments that incluse our selectedTreatment

KeeyanGhoreshi:
Addressing [DP2], not sure if it's what we want, but we'd still need to compute the powerset if I understand the combination thing correctly.  In your example, if we've selected surgery, we still need to compute the powerset of all treatments (minus surgery) and then add surgery back to the resultant sets.  Since it's possible to have 3 treatments, you could potentially have surgery/chemo/radiation, unless we're not including that.

Computing the full powerset is computationally expensive, but I can't think of another way to include combos that are more than 2 treatments.

Dtphelan1:
No that's a good point -- didn't think about that! A slight computational improvement (slight meaning not `O(n)` wise, but by a factor of `n-1`), would be to find the powerset of the combinations (not including the selectedTreatment), and afterwards append the selectedTreatment to each one of those possible sets. I don't think that's a priority right now, but the minimum I'd like to do would be to add a comment for future us in case that becomes needed.

KeeyanGhoreshi:
That's exactly how it works (it removes the selectedTreatment before computing the powerset) right now.  Even so, it's an exponentially increasing cost, O(2<sup>n</sup>), in the future if it needs to start computing `n` values that are large, we might want to look into bounding the length of possible sets (it is unlikely somebody has had 15 different types of treatments).

Dtphelan1:
Yeah we're on the same page then. Alright then I'm good on this front!

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
```diff
@@ -37,8 +37,8 @@ export default class TreatmentOptionsOutcomes extends Component {
 
         this.state = {
             similarPatientTreatments: this.generateSimilarPatientTreatments(props.similarPatients),
-            includedTreatments: [ 'surgery', 'radiation' ],
-            comparedTreatments: [ 'chemotherapy', 'hormonal' ],
+            includedTreatments: [ 'surgery' ],
+            comparedTreatments: [ 'chemotherapy', 'hormonal','radiation' ],
```
Dtphelan1: Should we give this a new name to reflect the combination-approach we take now?

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
```diff
@@ -173,14 +179,14 @@ export default class TreatmentOptionsOutcomes extends Component {
     renderRow(row, compareRow) {
         if (row.length === 0) return null;
 
-        const { name, totalPatients, sideEffects } = row;
+        const { name, displayName, totalPatients, sideEffects } = row;
```
Dtphelan1: [DP3] `name` is no longer used as a variable. Do we need to create this variable in the first place anymore? It's defined first in the `generateRow` function but i'm not sure we need it -- if we need to conditionally set displayName based on the contents of treatments, we can do that I think without the separate variable treatmentName.

KeeyanGhoreshi:
These changes need to be placed on top of Noranda's changes since they conflict.

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
KeeyanGhoreshi: I was being safe in case `name` was used elsewhere but I'll just remove `displayName` and turn `name` into `displayName`.

Dtphelan1:
Once DP3 is addressed, I'll approve. Good work 👍

src/mcode-pilot/components/TreatmentOptionsOutcomes/TreatmentOptionsOutcomes.jsx
```diff
@@ -134,9 +134,14 @@ export default class TreatmentOptionsOutcomes extends Component {
 
     generateRow(patients, treatments) {
         if (patients.length === 0) return [];
-
-        const treatmentName = _.isArray(treatments) ? treatments.map(name => TREATMENT_NAMES[name]).join(' & ') : TREATMENT_NAMES[treatments];
-        let row = this.initializeRow(treatmentName);
+        let displayName = _.isArray(treatments) ? 
+            treatments === this.state.includedTreatments ?
```
Dtphelan1: Are we looking for deep equality here? This will only return true if the variables are pointing to the same array; said another way, it will return false even if they have the same contents. If we want a shallow equality, we can use Lang.isEqual.

KeeyanGhoreshi: deep equality is not needed, in this case we actually want to check that they reference the same array, not the content.

Dtphelan1: Just wanted to check!

PR #602, opened 2019-03-15
[CCP] Integrate shim mapping functionality into flux notes

mgramigna:
For [Jira 1814](https://jira.mitre.org/browse/ASCODCP-1814).

For @dehall's original smart on FHIR datasource, only the McOde family could load in the app because the Synthea FHIR for these patients had MCODE profiles. This PR allows for Synthea patients not in the McOde family (i.e.without the profiles) to load into the app by passing them through a mapper.

I wrote the [fhir-mapper library](https://github.com/standardhealth/fhir-mapper) that does the mapping of Synthea FHIR to MCODE v0.5. This library allows for other mappers to be written in the future, e.g. when we need to write a mapper for Epic@BWH FHIR to MCODE. This library is now part of Flux Notes.

Also created `GenericSmartOnFhirDstu2DataSource.jsx` that extends the original Smart on FHIR datasource and implements the mapping of the generic Synthea FHIR to MCODE.

To test:

1. `yarn install`
2. Open the [SMART App Launcher](http://moonshot-dev.mitre.org:4001/?auth_error=&fhir_version_1=r2&fhir_version_2=r2&iss=&launch_ehr=1&launch_url=http%3A%2F%2Flocalhost%3A3000%2Flaunchsynthea&patient=&prov_skip_auth=1&provider=&pt_skip_auth=1&public_key=&sb=&sde=&sim_ehr=0&token_lifetime=15&user_pt=) and click "Launch App"
3. Select a patient *not from the McOde family*. This patient should now load in the app similar to how paitients in the McOde family do on master since the mapping is done
4. Feel free to also review the code on [fhir-mapper](https://github.com/standardhealth/fhir-mapper) that does the actual mapping

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/containers/SmartApp.jsx
+++ b/src/containers/SmartApp.jsx
@@ -28,7 +28,7 @@ class SmartApp extends FullApp {
                         this.dataAccess = new DataAccess("McodeV05SmartOnFhirDataSource");
                         this.loadPatient(smart.patient.id);
                     } else {
-                        this.dataAccess = new DataAccess(this.props.dataSource);
+                        this.dataAccess = new DataAccess(this.props.dataSource, this.props.dataSourceProps);
 
                         if (this.props.patientId) {
                             this.loadPatient(this.props.patientId);
--- a/src/dataaccess/GenericSmartOnFhirDstu2DataSource.jsx
+++ b/src/dataaccess/GenericSmartOnFhirDstu2DataSource.jsx
@@ -1,5 +1,5 @@
 import McodeV05SmartOnFhirDataSource from "./McodeV05SmartOnFhirDataSource";
-import processResources from './utils/fhir-entry-processor';
+import processFHIRResources from './utils/fhir-entry-processor';
 import mappers from 'fhir-mapper';
 
 class GenericSmartOnFhirDstu2DataSource extends McodeV05SmartOnFhirDataSource {
@@ -10,7 +10,7 @@ class GenericSmartOnFhirDstu2DataSource extends McodeV05SmartOnFhirDataSource {
 
     getPatient(id, callback) {
         super.fetchResources()
-            .then(resources => callback(processResources(resources, this._client.patient.id, this.mapper)));
+            .then(resources => callback(processFHIRResources(resources, this._client.patient.id, this.mapper)));
     }
 }
 
--- a/src/dataaccess/GenericSmartOnFhirDstu2DataSource.jsx
+++ b/src/dataaccess/GenericSmartOnFhirDstu2DataSource.jsx
@@ -1,7 +1,7 @@
 import McodeV05SmartOnFhirDataSource from "./McodeV05SmartOnFhirDataSource";
 import { syntheaToV05 } from 'fhir-mapper';
 
-class GenericMcodeV05SmartOnFhirDataSource extends McodeV05SmartOnFhirDataSource {
+class GenericSmartOnFhirDstu2DataSource extends McodeV05SmartOnFhirDataSource {
     constructor() {
         super();
         this.mapper = syntheaToV05;
@@ -12,4 +12,4 @@ class GenericMcodeV05SmartOnFhirDataSource extends McodeV05SmartOnFhirDataSource
     }
 }
 
-export default GenericMcodeV05SmartOnFhirDataSource;
+export default GenericSmartOnFhirDstu2DataSource;
--- a/src/dataaccess/GenericSmartOnFhirDstu2DataSource.jsx
+++ b/src/dataaccess/GenericSmartOnFhirDstu2DataSource.jsx
@@ -1,14 +1,16 @@
 import McodeV05SmartOnFhirDataSource from "./McodeV05SmartOnFhirDataSource";
-import { syntheaToV05 } from 'fhir-mapper';
+import processResources from './utils/fhir-entry-processor';
+import mappers from 'fhir-mapper';
 
 class GenericSmartOnFhirDstu2DataSource extends McodeV05SmartOnFhirDataSource {
-    constructor() {
+    constructor(props) {
         super();
-        this.mapper = syntheaToV05;
+        this.mapper = props && props.mapper ? mappers[props.mapper] : null;
     }
 
     getPatient(id, callback) {
-        return super.getPatient(id, callback, this.mapper);
+        super.fetchResources()
+            .then(resources => callback(processResources(resources, this._client.patient.id, this.mapper)));
     }
 }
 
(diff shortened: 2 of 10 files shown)


src/dataaccess/McodeV05SmartOnFhirDataSource.jsx
```diff
@@ -105,6 +105,13 @@ class McodeV05SmartOnFhirDataSource extends IDataSource {
                 // response is a Bundle of type searchset
                 // TODO: error handling?
                 if (response && response.entry) {
+                    // If we need to use a mapper to add profiles to the fhir, update the resources accordingly
+                    if (mapper) {
+                        const resources = response.entry.map(e => e.resource);
+                        const results = mapper.execute(resources);
+                        const wrappedResults = results.map(resource => ({ fullUrl: `urn:uuid:${resource.id}`, resource, request: { method: 'POST', url: resource.resourceType } }));
```
dehall: I know this was in the sample code that I sent to you, but looking at it now I'm concerned about rewriting the fullUrl like this. Maybe it doesn't matter though. Making a note here mostly so I remember to look at this again on Monday

dehall: Do you see any good way of preserving the fullUrl from the entry through this mapping process? I think the mapper will preserve order but I don't know if it's guaranteed.

When I originally wrote this, it was in the context of updating local json bundles generated by synthea, and I know that this was how the fullUrl was originally created, with resource.id being a UUID so it's globally unique. Arbitrary data coming from a server may have resource IDs that are unique among that resource type but not globally unique. So my fear is that this may produce entries with duplicate fullIUrls.

gregquinn2001:
[GQ1] All data sources provide Flux Notes Object Model as output and accept it as input. What differentiates them from each other is what data format and protocol they use to receive their data. The name of the data source should reflect the expected input format and protocol. In this case, the data source supports generic FHIR DSTU2 I believe, right? I think we should rename the data source to reflect that.

mgramigna:
> [GQ1] All data sources provide Flux Notes Object Model as output and accept it as input. from UI. What differentiates them from each other is what data format and protocol they use on the other side to retrieve and store data external to the application. The name of the data source should reflect the expected format and protocol for data between the application and the external data source. In this case, the data source supports generic FHIR DSTU2 I believe, right? I think we should rename the data source to reflect that.

Makes sense! Would `FHIRDSTU2SmartOnFhirDataSource` make sense?

gregquinn2001:
For [GQ1], that sounds good but maybe FHIR in there only once so GenericSmartOnFhirDstu2DataSource?

mgramigna:
@gregquinn2001 Sounds good. I also added your blurb about naming convention to the [data source wiki page](https://github.com/FluxNotes/flux/wiki/Flux-Notes-Data-Sources#naming-convention)

mgramigna:
Resolved GQ1 and rebased. Will look into Dylan's comment

mgramigna:
[MG1/GQ1] From Greg: mapper should be specified at the metadata level rather than passed in from the subclass

mgramigna:
Resolved MG1/GQ2. `McodeV05SmartOnFhirDataSource` now knows nothing about the mapping. Mapper is specified at the `AppManager` level and passed down accordingly.

There is now a separate app for Synthea FHIR without MCODE whose launch url is now `http://localhost:3000/launchsynthea`. I've updated the PR description for testing accordingly

src/dataaccess/McodeV05SmartOnFhirDataSource.jsx
```diff
@@ -1,6 +1,5 @@
 import IDataSource from './IDataSource';
-import PatientRecord from '../patient/PatientRecord';
-import ObjectFactory from '../model/ObjectFactory';
+import processResources from './utils/fhir-entry-processor';
```
gregquinn2001: One minor nit! Can we name processResources to processFhirResources since it specifically processes FHIR resources? Looks really good though! Thanks!

mgramigna:
Resolved the renaming.

[MG2] Test is failing because it expects non-profiled FHIR to error out but it no longer errors out :). Update this test to test mapping functionality with new datasource instead

gregquinn2001:
Good catch! I was just about to run the tests actually because I had forgotten to do it

mgramigna:
Tests should pass now, including functionality to test that the mapping from synthea => mcode v0.5 works

mgramigna:
All outstanding comments should be resolved.

New patch of `fhir-mapper` partially addresses @dehall's comment, so will need to re-run `yarn install`

src/containers/SmartApp.jsx
```diff
@@ -28,7 +28,7 @@ class SmartApp extends FullApp {
                         this.dataAccess = new DataAccess("McodeV05SmartOnFhirDataSource");
                         this.loadPatient(smart.patient.id);
                     } else {
-                        this.dataAccess = new DataAccess(this.props.dataSource);
+                        this.dataAccess = new DataAccess(this.props.dataSource, this.props.dataSourceProps);
```
dehall: If `dataSourceProps` is going to be a thing, should it be added to the other call in the FullApp and CoreCancerPilotApp as well? Or it's fine with me to save that until it would actually be used there.

gregquinn2001: Yeah I thought about that too. I think it's fine to only pass it into the data sources that use it because that can be known in the factory method.

src/dataaccess/GenericSmartOnFhirDstu2DataSource.jsx
```diff
@@ -0,0 +1,17 @@
+import McodeV05SmartOnFhirDataSource from "./McodeV05SmartOnFhirDataSource";
+import processFHIRResources from './utils/fhir-entry-processor';
+import mappers from 'fhir-mapper';
+
+class GenericSmartOnFhirDstu2DataSource extends McodeV05SmartOnFhirDataSource {
+    constructor(props) {
+        super();
+        this.mapper = props && props.mapper ? mappers[props.mapper] : null;
```
dehall: For posterity: `props.mapper` is a string so this can only reference mappings that are already defined in the fhir-mapper library. I was about to propose making this more generic to support defining arbitrary mappings in the props, but then I realized if we have a need to create a specific mapping, then we could just define the mapping in a new DataSource, ex "SpecificEhrVendorSmartOnFhirDataSource", so I'm fine with how this works now.

dehall:
FYI I'm going to test this out with a different SMART sandbox at https://sandbox.hspconsortium.org just so we aren't all using the exact same data to test it with. HSPC sandboxes also use Synthea-generated patients so it should look pretty similar, but there may be differences we should be aware of.

dehall:
Btw it worked with the different sandbox: 
hspc

gregquinn2001:
Very cool! That demonstrates the option we need to implement in launch parameters for smart on fhir to suppress our banner in case the EHR provides one. I think I added a iira for that a while ago but I'll make sure and move it to the top of the pile for evaluation for next sprint! Thanks for testing!

PR #615, opened 2019-03-28
Update context insertion to properly reflect location in editor

mgramigna:
See [Jira 1850](https://jira.mitre.org/browse/ASCODCP-1850) for screenshots of the original bug. When a context was added at the beginning of the note, it would still appear last in the active context breadcrumb. This was due to the implementation of `addShortcutToContext` incorrectly detecting where to splice the context.

I changed the implementation to allow for two possible options for knowing where to insert the context:

1) There is now a property on `Context` objects called `initialContextPosition` that specifies where to splice it relative to other contexts
2) Make a call to `FluxNotesEditor` to get all the context shortcuts before the cursor position, and splice in the new context after those

These are the following ways to add context to a note (I think I got them all):

* Typing the shortcut
    * Hitting enter or clicking on portal suggestion
    * Typing out the full shortcut without clicking/enter
* Inserting a context from the TDP into a note
* Pasting a context into a note
* Loading a close in-progress note
* Inserting a template that has context shortcute
* Inserting a context from the context tray
* Flux commands for inserting structured phrases

All of these cases should result in the context being inserted in the correct position, which will be seen in the active context breadcrumb.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/context/Context.jsx
+++ b/src/context/Context.jsx
@@ -2,7 +2,8 @@ import Lang from 'lodash';
 
 export default class Context {
 	constructor() {
-		this.children = [];
+        this.children = [];
+        this._initialContextPosition = -1; // Where to insert the context relative to the others on creation
 	}
 
 	initialize(contextManager, trigger = undefined, updatePatient = true) {
@@ -117,5 +118,13 @@ export default class Context {
 
     isGlobalContext() {
         return false;
+    }
+
+    get initialContextPosition() {
+        return this._initialContextPosition;
+    }
+
+    set initialContextPosition(initialContextPosition) {
+        this._initialContextPosition = initialContextPosition;
     }
 }
--- a/src/context/ContextManager.jsx
+++ b/src/context/ContextManager.jsx
@@ -214,15 +214,6 @@ class ContextManager {
         this.activeContexts = [];
         this.contextUpdated();
     }
-
-    // Clears all non active contexts from this.contexts
-    // Only used after picking options from template
-    // When choosing options, many unncessary contexts get added
-    clearNonActiveContexts() {
-        this.contexts = this.contexts.filter((context) => {
-            return this.activeContexts.includes(context);
-        });
-    }
 }
 
 export default ContextManager;
--- a/src/context/ContextManager.jsx
+++ b/src/context/ContextManager.jsx
@@ -41,6 +41,10 @@ class ContextManager {
         this.activeContexts = contextsToKeep;
     }
 
+    setGetContextsBeforeSelection(getContextsBeforeSelection) {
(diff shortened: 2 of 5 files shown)


jafeltra:
This looks good! I came across two things. We can evaluate if either are something we want to actually fix now.

[JA1] Inserting the "test" template sets the context correct. However, if I add #staging from the context tray at the bottom of the note, the order of the contexts is wrong. It has context as "GIST, staging, Fracture."

[JA2] When using the "test" template, I've been getting the error at the bottom of the editor saying "GIST cannot be deleted because oncohist depends on it" quite often as soon as I try to type right after inserting the template. I don't know if this is related to context ordering or not. I can sometimes get it to happen on fluxnotes.org too. It doesn't seem to matter what order I add the conditions in the template because I've gotten it for both cases. However, I don't get it every time I add the template and I haven't figured out when I do and when I don't. (Sorry I don't have a better way to replicate yet.)

Dtphelan1:
[DP1] Weird case, also to do with templates: If I have local contexts in the editor before I insert a template, after I finish inserting the template they won't appropriately be calculated as a part of context _until_ I close and reopen the note.

Dtphelan1:
N.B. DP1 also happens on Master

mgramigna:
Looking into JA1 now

Per JA2, there seems to be a bug where the GIST shortcut will stay bolded in the template even after it is selected/the template is inserted. In the case where it's bolded like this, that is when the error shows up. Hard to replicate, but the same behavior is on fnorg. I think it has to do with how quickly you slide the mouse down to the OK button for the test template...

jafeltra:
Thanks for the update on JA2. I didn't realize that's what was happening. I think you're good to ignore JA2 in that case.

mgramigna:
Resolved JA1 and DL1. 

Let's maybe hold off on JA2 for now, especially considering that template insertion might be getting an overhaul during the 13 pointer.

gregquinn2001:
How are we ruling JA2 out of scope? It sounds like it is related to the context issue you were solving???

mgramigna:
@gregquinn2001 It's a consequence of the bug where the mouse out event may sometimes not trigger properly if you move your mouse too quickly to the OK button when inserting the "test" template.

gregquinn2001:
Ok. Sounds good then. The OK button will definitely be gone with the new note flow changes so the bug should be gone with it it sounds like.

Dtphelan1:
DP1 is addressed! Just giving a once over before approving! 👍

Dtphelan1:
Looks good to me! 😄

PR #651, opened 2019-05-17
Tablet mode toggle

sgober:

## Submitter:
JIRA 1964 for the tablet toggle between the TDP and POC.  See the JIRA for mockups.

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


src/dashboard/PointOfCareDashboard.jsx
```diff
@@ -13,17 +23,98 @@ export default class PointOfCareDashboard extends Component {
...
         };
+        const PanelComponent = this.state.showPOC ? PointOfCarePanel : TargetedDataPanel;
 
         return (
             <div id="point-of-care-dashboard-content" style={{display: "flex"}}>
-                <div className="right-border-box" style={targetedDataPanelStyles}>
-                    <TargetedDataPanel
+                {this.renderSidebar()}
+                <div className="right-border-box" style={panelStyles}>
+                    <PanelComponent
```
jafeltra: This is definitely a question, but the way we have this switch between components set up looks like it means we'll have to pass all the same props into PointOfCarePanel and TargetedDataPanel. I'm not exactly sure everything that the PointOfCarePanel will need to do, but it seems like there could be a chance one component could need props that the other would not. Is it better to just conditionally render the different components? Or is it okay to pass the props and not use them in the components? Or maybe the props will end up being very similar and this is the best way to do it. Just my thoughts that I wanted to mention.

jafeltra:
This looks good! The sidebar looks really cool! A few things I noticed:

[JA1] I think the icons might be swapped. In the mock ups it looks like the lower icon with the 6 boxes is for the POC and the upper icon with the note is for the TDP
[JA2] Do we need that little white arrow that's in the mockups indicating which tab you have selected in the sidebar?
[JA3] The right side of the TDP gets cut off for me sometimes. It happens when I initially load the app, when you switch between POC and TDP, and in tablet portrait mode on the iPad view. It seems like this might be due to the sidebar.
[JA4] This is more of a question, but is there a reason to have the svgs in the public folder if we are adding the svg directly in the component?

src/dashboard/PointOfCareDashboard.jsx
gregquinn2001: @jafeltra yes! This is not a planned point of extensibility per se so we should have a state variable that we use to conditionally render either TDP or POC

sgober: @jafeltra @gregquinn2001  you mean have separate render functions for rendering the TDP and POC?  I was thinking about doing that but didn't in case the inputs were similar to the POC  component.  I'm okay either way though, someone make a decision!

sgober:
@jafeltra 

[JA1] you would be correct... fixed
[JA2] not sure... it was not provided with the icons that dan sent. @gregquinn2001 what do you think?
[JA3] I can't recreate this on a normal screen but am seeing this for iPad mode.  looking into it
[JA4] I followed the convention of what we do for other icons.  I can remove them but it seems nice to have them located in the repo.

src/dashboard/PointOfCareDashboard.jsx
gregquinn2001: It doesn't need to be separate render functions IMO. I think it should just be a switch statement on the state variables that says whether we are displaying POC or TDP and in the POC case, it returns <PointOfCare.... and in the TDP case it returns <TargetedDataPanel ...

It doesn't need to treat them as the same thing with a common abstract base. They are different other than both being react components. Does that make more sense?

gregquinn2001:
For JA2, I do think we want it there eventually. We can make a separate JIRA for it and ask Dan for it so we can add it in a future sprint

gregquinn2001:
For JA4, we don't use the SVGs at all in the public folder? I'd say don't include them there then? Other thoughts?

jafeltra:
[JA5] The icons look like they're used for the correct display now. But I did notice that in the mockups, they TDP icon is above the POC icon. Also, it looks like in the code, the variables for the `TdpIcons` are actually the icons for the POC and the `PocIcons` variables are the TDP icons. We might want to swa the names of those just so we don't get confused in the future.

jafeltra:
I was also talking to @Dtphelan1 about JA4 and I think we reached the same conclusion as @gregquinn2001 mentioned. I think we should remove the SVGs from the public folder since we don't seem to be using them at all from there.

sgober:
JA3 & JA4 resolved, also addressed the separate render comment

jafeltra:
JA1, JA3, and JA5 all look good to me. JA2 and JA5 are still outstanding I believe.

sgober:
JA5 resolved

src/dashboard/PointOfCareDashboard.jsx
jafeltra: The change for this looks good! Thanks!

PR #654, opened 2019-05-22
Support completing attribute shortcut labels via Context Portal

mgramigna:

Can now complete attribute labels (e.g. #status). Currently, the functionality works similar to something like @condition. You can type #status in the editor and then the picker portal will appear to select a value. I did this by adding a picker option to the shortcuts in `Shortcuts.json` as well as update the triggers we manually set in `ShortcutManager` to use said picker.

Things not covered that will be covered by other tasks:

* Completion of attributes that support multiple options (e.g. #reasons)
* Ability to close the context portal and re-open for selection when your cursor is next to an incomplete attribute.

Test case:

1) type #disease status (should show up as incomplete)
2) type #status (#disease status should now be complete, #status should be incomplete and context portal should open)
3) Select a value, should replace #status with the selected value.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


jafeltra:
This looks cool! One thing I noticed:

[JA1]: When I type #disease status #reasons, before picking a choice, reasons doesn't show up as incomplete anymore. It does on the base branch.

sgober:
[SO1] When selecting a reason, I am seeing a rogue comma show up in the TDP before the reason I selected.

mgramigna:
Resolved JA1 and SO1. They were the same root cause, and I was able to just delete the calls to `setText(<initiating trigger>)` since we don't need to do that anymore after @karenzheung 's refactor. This was causing the attribute value for #reasons to be set prematurely.

mgramigna:
Re: JA2

This happens because of the call to `shortcut.onBeforeDeleted()` on line 402 of `FluxNotesEditor`, which has been in there for a while. This gets triggered when you hit Esc to close the context portal when it is open. This only used to happen for Inserters, but now that we have a picker for CreatorChild shortcuts, we go into this logic too.

Part of this logic removes the shortcut that you are selecting a value for from its parent context's array of children. This reason it removes the shortcut from the children when you hit Esc is because it is added to that array of children every time `newCurrentShortcut` is called in `FullApp` which is called when selecting from the suggestion portal before the context portal even opens.

Now that we actually has code that does something with `this.children` for something like #disease status, we encounter a case where we don't want `onBeforeDeleted` to be called for something like #status because we still want it to be in #disease status' array of children even if it doesn't have a value, as that is what determines the incompleteness of #disease status. We can't just remove it though, because we can then encounter a case where if act crazy enough you could add infinitely many shortcuts to `this.children` if you repeatedly choose from the suggestion portal and then immediately close the Context Portal.

I think this might be more of an issue with the `shortcut-entry-overhaul` branch as a whole rather than this PR, but I am open to more discussion about this tomorrow.

**TL;DR: Existing behavior is causing this that we never caught before we had a label approach, might want to think about resolving it outside of this PR instead of a JA**

Dtphelan1:
I think the suggestion to make this a separate JIRA makes sense to me. The root problem exists on the overhaul branch separately from this, will impact other features beyond just this, and falls outside the immediate scope of this functionality. That's my two sense.

jafeltra:
I agree with both comments, and I think it makes sense to add a Jira that will be part of the Shortcut Guidance epic since it's not directly related to this PR/task.

PR #657, opened 2019-05-24
Mock data update

KeeyanGhoreshi:
This PR updates the `mock-data` for the compass app to use coded values for the appropriate fields.  This really only affects the treatments, side effects, TNM staging, and tumor markers.  The mock data updater was used to replace simple string keys in the mock data with coded values.  

The internal model was then updated to support the coded values, so replacing lookup tables with the coded values `displayName` and using the code/codeSystem to identify the values to match when finding similar patients and filtering.  The only hiccup in this transition is the usage of the checkboxes, which cannot take Objects, so the coded values had to be broken into a unique string that could then be selected by a checkbox and used to identify the correct codified value.  The solution I came up with is a bit awkward, appending the code and code system together into a single string to make a unique identifier.  

Issues can also arise if a coded value doesn't have a displayName or a codeSystem, but the data received by the app is already preprocessed to some degree, so it doesn't seem likely that the filter function will have to deal with coded values missing fields.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [ ] Existing tests passed

## Reviewers:

- Try to break the code


Dtphelan1:
I think the "codeSystem + codeValue" as unique identifier is pretty consistent with what a) we've done in other places, and b) What most of those codeSystems themselves suggest to do. 

On the whole I think this looks fine. Nothing breaks, and we do now use something that resembles a coded value. A few minor concerns: a) should we be separating `names`, the array, from the `displayName` on each value object; b) are we defining our own cardinality system here with "multi-choice" and "choice"? 

If we can chat about those comments a bit, I think I'll be good to approve. Interested in hearing what @dehall has to add on the matter.

src/mcode-pilot/mock-data/transformedDataModel.json
```diff
@@ -2,33 +2,114 @@
     "demographics":{
         "birthDate": {"type": "date", "values": [1920,1994], "weight": "uniform"},
         "gender": {"type": "choice", "values": ["male","female"], "weight": "uniform"},
-        "race": {"type": "choice", "values": ["White","Black","Unknown","Other unspecified (1991+)","Other (American Indian/AK Native; Asian/Pacific Islander)"],
-                                   "weight": [0.5,     0.2,    0.05,     0.01,                       0.24]}
+        "race": {"type": "choice", "values": [{"code":"2106-3", "codeSystem":"http://hl7.org/fhir/v3/Race", "displayName": "White", "codeSystemName":"cdc"},
```
rdingwell: Just want to make sure I am correct in assuming that this file was only used for updating the mock data, it's not used for anything in the running application, correct?

src/mcode-pilot/utils/filterTreatmentData.js
```diff
@@ -62,29 +41,29 @@ function generateTreatmentData(similarPatients, treatments, includedTreatments)
     if (similarPatients.length === 0) return [];
     let treatmentData = [];
     treatments.forEach(treatment => {
-        const filteredPatients = similarPatients.filter(patient => isSame(patient.treatments, treatment));
+        const filteredPatients = similarPatients.filter(patient => isSame(patient.treatments.map((e) => { return typeof e === 'object' ?e.code : e; }), treatment.map(treat => { return treat.code; })));
         let displayName =
             _.isArray(treatment)
                 ? (treatment === includedTreatments)
-                    ? treatment.map(name => TREATMENT_NAMES[name]).join(' & ')
+                    ? treatment.map(name => name.displayName).join(' & ')
```
rdingwell: Seeing how name is no longer a sting/name can we change the variable name to something more appropriate

src/mcode-pilot/utils/filterTreatmentData.js
```diff
@@ -103,7 +82,12 @@ function generateSimilarPatientTreatments(similarPatients) {
 
     similarPatients.forEach(({ treatments }) => {
         treatments.forEach(treatment => {
-            similarPatientTreatments[treatment] = { key: treatment, name: TREATMENT_NAMES[treatment] };
+            if (typeof treatment === "object") {
+                similarPatientTreatments[treatment.code] = { key: `${treatment.code},${treatment.codeSystem}`, name: treatment.displayName, reference: treatment};
```
rdingwell: shouldn't the key and the similarPatientTreatments key be the same.  There is a small chance of code being applicable across codesystems which would be an issue

src/mcode-pilot/utils/filterTreatmentData.js
```diff
@@ -149,23 +140,20 @@ function isSimilarPatient(treatmentDataPatient, similarPatientProps) {
                 // pathology
                 } else if (mcodeElement === 'onco.core.TumorMarkerTest') {
                     const receptorType = reference._tumorMarker._findingTopicCode.codeableConcept.coding[0].code.value;
-                    if (receptorType === '16112-5' && (!tumorMarkers.er || _.lowerCase(tumorMarkers.er) !== _.lowerCase(value))) {
+                    if (receptorType === '16112-5' && (!tumorMarkersLabeled["16112-5"] || tumorMarkersLabeled["16112-5"].value.code !== reference.findingResult._value._coding[0]._code.code)) {
```
rdingwell: This should be a generic capability that can be used across tumormarkers without the need of hard coding values for specific types

src/reducers/initial.json
```diff
@@ -2,9 +2,11 @@
     "totalPatients": 0,
     "totalSimilarPatients": 0,
     "similarPatientTreatments": [],
-    "includedTreatments": ["radiation"],
+    "includedTreatments": [{"code":"385798007", "codeSystem":"http://snomed.info/sct", "displayName": "radiation therapy", "codeSystemName":"SNOMED"}],
```
rdingwell: I'm not certain but I think @dehall deleted this file.  If not clear all of the values for the treatments, the values this file provides are also derived in the record to props functionality.

dehall: I didn't delete the file itself, I think it's required for redux. I did clear out the values and just left an empty object so like you said the same thing should work here.

src/mcode-pilot/mock-data/transformedDataModel.json
KeeyanGhoreshi: Yes this file is strictly for updating the mock data.

Dtphelan1:
Once my comments get a brief mention and Rob's feedback gets addressed I'll review again @KeeyanGhoreshi

KeeyanGhoreshi:
Recent commit should address all of Rob's comments.  As for Dylan's comments:

(a) The `names` array corresponds to the current mock data's string values and helped the script replace them with the correct coded value.  For something like the side effects, the display name might be "reduced libido", but its name in the `names` array would be `decLibido` to align with the old mock data's representation of that side effect. 

(b) The `choice` and `multi-choice` does define something like a cardinality system, but its strictly part of the data model which was used to replace the mock data.  Outside of the script that updates the mock data, it doesn't mean anything.

Dtphelan1:
Given the limited scope of the date model I made comments on, I'm fine with the things I asked for clarification on. I'll do another pass through and then aget back with my updated review.

src/mcode-pilot/utils/recordToProps.js
```diff
@@ -1,9 +1,13 @@
 import _ from 'lodash';
 import FluxTumorDimensions from '../../model/tumor/FluxTumorDimensions';
+import FluxTNMClinicalPrimaryTumorClassification from '../../model/oncocore/FluxTNMClinicalPrimaryTumorClassification';
+import FluxTNMClinicalRegionalNodesClassification from '../../model/oncocore/FluxTNMClinicalRegionalNodesClassification';
+import FluxTNMClinicalDistantMetastasesClassification from '../../model/oncocore/FluxTNMClinicalDistantMetastasesClassification';
 
 export default function getProps(patient, condition) {
 
     const tumorMarkers = patient.getMostRecentTumorMarkers(condition);
+    const tnminfo = processPanel(condition.getMostRecentClinicalStaging()._tnmStageGroup._panelMembers.observation, patient);
```
rdingwell: If there isn't a most recent clinical staging this line is going to cause runtime errors

rdingwell:
Travis is failing on the latest change for some reason.  I restarted the builds but they are still failing. Once the PR passes travis I will approve.

Dtphelan1:
This looks good to me. Good work Keeyan!

PR #665, opened 2019-05-31
Completion Portal Plugin and Completion Component Separation

Dtphelan1:
This PR includes Sarah's task, to separate the completion components (ContextCalendar and ContextListOptions) from the EditorPortal component, and my work separating the ContextPortal into it's own plugin. As a product of my work, a few other things have been done: 
- I've removed old ContextPortal code, renamed old functions and modified/removed code that supported that previous approach from FluxNotesEditor. If you find remaining code related to our old approach, flag it for removal. 
- We need to ensure SuggestionPlugin doesn't try to match against that shortcut text. In order to accomplish this, we've added a new option to SuggestionPortalPlugin, `typesToIgnore`, which is an array of types of nodes to ignore when matching text. This ensures that our SuggestionPlugin isn't matching against structured text. 
- I've improved the math for switching menu-index in the contextListOptions. Then moved the new index-selection math into suggestion portal's code. This will work better on long lists. 

Existing shortcomings that I'm working on: 
- Date selector loses editor focus after selection of a value.

Additionally, there are some bugs that were uncovered as a part of the development of this task, but are unrelated/should be separate JIRA's I think: 
- When selecting a date for a new shortcut, the value in the editor shows up as `undefined`. This seems to be introduced/exacerbated as a part of Matt's recently merged Attribute-selection branch. We need a JIRA to address this. 
- Similarly problematic, but the value of the date prior to selection invalid. This can be seen for example when filling out disease status. Select a value (e.g. stable) and then watch the TDP when you open the as of date picker. The display is 'invalid date", which I claim is a result of calling the moment.js format function on an improperly formatted date value. 

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [ ] Existing tests passed

## Reviewers:

- Try to break the code


sgober:
NOTE: `existing-shortcuts-refactor` will need to be deleted as well once this is merged.

Also, I can't review this one since I worked on it with @Dtphelan1.

mgramigna:
A few things that I found. At least one exists on our base `shortcut-entry-overhaul` branch, but I will note them here just so we are aware, and indicate as such.

[MG1] Hitting Escape to close the portal causes this console error: 
image

[MG2] This branch doesn't have the functionality for completing attribute shortcuts that was introduced in #654. I believe this is because of line 189 of `CreatorChild.jsx`. The completion component should be a `ContextListOptions` if the picker is not "date-id", as this is what we use to complete attribute shortcuts. But it is returning null.

[MG3] I can now delete the characters of inserters 1 by 1 if I do any of the following:

* Insert @condition so it shows up as incomplete, but leave the ContextListOptions completion component open without selecting a value. Hit backspace a few times, and you can delete the characters 1 by 1 (I think we've seen this when experimenting with completing @conditions before)
* Insert @condition so it shows up as incomplete, but hit Escape before selecting a value.
    * Hit space to continue typing, and then hit left arrow to bring the cursor back to right after the incomplete so the portal re-appears
    * When the portal reappears, select a value
    * You can now delete the characters of that inserter 1 by 1 using backspace
* Insert @patient or any other inserter that doesn't require you to select an option from the portal afterwards. You can delete the characters 1 by 1 using backspace

[MG4] **This is on the base branch but I found it here = JIRA** If I author a note with #disease status or any other hashtag and then sign the note, upon re-opening the note, the hashtags still appear with the dotted underline style when they should be solid underlined since the note is signed.

Dtphelan1:
MG1 - Yeah, good catch. I noticed that too, and will look into if there's a better way we can exit the portals. Fundamentally I think this has to do with the fact that we don't really use the React portal's properly (note: either does suggestion portal). That compounded by the fact that this isn't a breaking bug makes me think we should triage the severity of this if it's not an easy fix. 

Addressed MG2! 

MG3 - This is very odd. I haven't touched anything on shortcuts except adding a CompletionComponent getter. While some of these might be explained by peculiar behavior resulting from the menu being open, the @patient doesn't obviously relate to anything I touched...  I'll do some digging

MG4 - I think a JIRA sounds right here.

Dtphelan1:
Got to the bottom of the MG3 bug: based on the changes Sarah made, every Inserter and CreatorChild has a completionComponent by default, 

We only need certain shortcuts to have defined completionComponents. Will talk with @gregquinn2001 about what is the ideal way to accomplish this; one possibility is that we define in metadata a string that corresponds to the completionComponent to be used by the shortcut, and the class (either a base class or one of the subclasses) could define the getter that translates that metadata string into the completionComponent.

mgramigna:
I agree with all the above points. I'll pull down the most recent stuff and continue reviewing!

Dtphelan1:
MG3 should be addressed! :)

Dtphelan1:
General update: I've been looking into the bug I mentioned, which relates to the date-picker component `rc-calendar`. Specifically, the issue is that selecting a value with the enter key causes the editor to lose focus in cases where there is no text after the shortcut `#date`. Selection is *not* lost if either: 
- The editor has some text after `#date`
- The value was selected with a mouseClick. 

I've been looking into what might be happening with the selection in these cases, but I can't figure out from first principles what's happening differently in the case where there is no text after our date-picker and the user uses the enter key.

danlee1025:
[DL1]  Found a bug where two portals are open.
- Create a new note
- Enter @condition, escape out of condition selection.
- Hit left arrow key until you get to beginning of line and then hit right arrow key twice.

image

sgober:
I'm seeing some different behavior with the menus now, not that they necessarily should be fixed, more discuss what the expected behavior is.  See the below screenshots:
Screen Shot 2019-06-06 at 10 41 04 AM
Screen Shot 2019-06-06 at 10 41 44 AM

Dtphelan1:
@sgober Great find - this is not a bug that has to do with what's been changed here; this is an issue with the getPosition function we pass to components from the FNE. We should create a JIRA and fix that

sgober:
@Dtphelan1 I noticed it when working on the get help menu, so that is affected as well.  Can you handle making the JIRA?

Dtphelan1:
### The gift that keeps on giving

The Good: I've successfully fixed the date-picker bug I've been investigating. 

The Bad: I've done it in a terrible way. After days of investigating, I've been unable to determine why the bug is happening. What I have determined though is if I insert a space and then delete it immediately before returning our transform, the date selector seems to work properly. This is not good practice. That said, there are other bugs that need addressing if this task is going to move forward. I'm creating a JIRA to address this bug in the course of our overhauling shortcuts, where I'll capture as many of my learnings as best as I can. 

The Additional Bad: I think I found a [DP2]; our editor doesn't seem to gain focus after insertion from the right hand side. Have yet to test shortcutEntryOverhaul, so I'm not sure if this is a true DP2. 

Appendix: Some of the factors that contribute to the difficulty here are: 
- Our use of focus and blur in multiple places around the application makes it difficult to figure out what
- Slate JS being broken up into so many files make tracking function calls across the app less-than straightforward. 
- The mix of date-picker (a non-plugin component) into a largely plugin-mediated stack of event handlers. 
- (possibly) Our use of string refs, a practice that has been phased out in more recent versions of react in favor for callback refs and explicit ref instantiation. Not sure if this is biting us, but it's a possible loose-end. See here for more details on string refs being deprecated: https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

Let me know if you ahve any other questions about what I've done or what I've tried so far. Please let me know if you have any ideas or suggestions on what I could suggest. And pray for me.

Dtphelan1:
Mediocre News: The DP2 shows up on `shortcut-entry-overhaul`, which means I don't think it should fall under the scope of this task. I'll be making a JIRA now.

Dtphelan1:
Confirming that DL1 has been identified as a bug separate from this PR. 

Additionally, MG1 is not an easy fix/is also somewhat separate. It has more to do with the somewhat improper usage (started in suggestion portal, carried over into our portals) of the react-portal component. I'll create a task in the backlog, but it shouldn't be a terribly high priority for now. 

With that, I think all existing issues have been resolved/separated/triaged appropriately. Review round 2? 👍

danlee1025:
I think this looks good now, just needs a rebase.

mgramigna:
General question, should we be able to come back to a selected value and re-select it with this plugin approach? I don't see that functionality right now, but I wasn't sure if we planned on making that a separate Jira to begin with.

CC @gregquinn2001

gregquinn2001:
Yes the user should be able to go back to stable for example it should pop up completion portal with stable selected and the user change the value regardless of how stable (aka #stable) got into the note! So if type #stable in a note then I should be able to put my cursor next to it and have it pop up the ability to select any of the status values. I don't think Dylan has addressed that so it should be a separate task/subtask somewhere if it isn't already.

Dtphelan1:
As @mgramigna points out, my code supports exactly this behavior but was held up by the fact that previous workflows removed all the valueOptionsToSelectFrom once one was selected. I will connect with Matt on the ~one-line fix and test appropriately to confirm nothin wonky happens.

gregquinn2001:
It may be simple but it should be a separate JIRA! It was not part of this task.

mgramigna:
Re-read through everything and re-reviewed. All corresponding Jiras have been made, and I haven't found any new concerns.

This needs one more rebase, which will incorporate the new shortcut model changes from #666. Hopefully it's not too bad!

src/notes/FluxNotesEditor.jsx
```diff
@@ -1913,18 +1937,20 @@ class FluxNotesEditor extends React.Component {
...
-                        contextManager={this.contextManager}
-                        contexts={this.state.portalOptions}
-                        getPosition={this.getTextCursorPosition}
-                        onChange={this.onChange}
-                        openedPortal={this.state.openedPortal}
-                        onSelected={this.onPortalSelection}
-                        state={this.state.state}
-                        trigger={"@"}
-                    />
+                    {CompletionComponent &&
```
gregquinn2001: Why do we need to expose the completion component to FNE at all? Why can't we just render the completion component within the render method of CompletionPortal?

src/shortcuts/CreatorChild.jsx
```diff
@@ -70,7 +70,7 @@ export default class CreatorChild extends Shortcut {
         let result = super.onBeforeDeleted();
         if (result && !Lang.isUndefined(this.parentContext)) {
             const parentAttributeName = this.metadata.parentAttribute;
-            if (this.metadata["subtype"] && this.metadata["subtype"] === "list") {
+            if (this.metadata["subtype"] && this.metadata["subtype"] === "multi-choice") {
```
gregquinn2001: Awesome! :)

src/shortcuts/Shortcut.jsx
```diff
@@ -171,6 +173,24 @@ class Shortcut extends Context {
     get isComplete() {
         return true;
     }
+
```
gregquinn2001: I don't think Shortcut model class should be responsible for creating a completion component. Let's have a separate factory method that is called from CompletionPortal and provides the shortcut or just its metadata as an argument and returns the correct completion component

src/notes/FluxNotesEditor.jsx
```diff
@@ -67,6 +68,31 @@ const initialEditorState = {
...
+        // Clean up some variables stored at the Editor level
+        this.setState({
+            completionComponentShortcut: null,
+            portalOptions: [],
+            completionComponent: null,
+            openedPortal: null,
+        });
+    }
+
+    getCompletionComponent = () => {
```
gregquinn2001: I'd like to see this method go away

Dtphelan1: So could we talk about renaming the function? We need a way to get a direct reference to the component so we can access the onKeyDown function, so we can't get read of this component per se.

gregquinn2001: I'm looking at it more now. My comment was not about the name; it was more about that I'd like to see us not need that function.... maybe that's wishful thinking.

gregquinn2001: Can we add a render method to the CompletionPortalPlugin that renders and can provide a reference to the completion component?

mgramigna:
I'm assigning @gregquinn2001 to this, seeing as Dan will be out and Greg has raised some questions that need addressing so we'll require both our approvals before merge.

src/notes/FluxNotesEditor.jsx
Dtphelan1: Originally the short answer seemed to be no - the primary reason that I previously thought the component needed to be rendered at the FNE level was because FNE needs to give CompletionPortalPlugin some way to reference the `ref` for the actual completionComponent. I now think that this can be done using some slightly more advanced react code, so I will spend some time tomorrow morning looking into this.

src/shortcuts/Shortcut.jsx
Dtphelan1: Four important things need to happen w/r/t completionComponents: 
- the CompletionPortalPlugin needs to be able to see whether or not a given shortcut we're neighboring has a completion component; 
- our CompletionPortal needs a reference to the completionComponent it is responsible for rendering; 
- our plugin needs a reference to the DOM node of he completion component so we can use that objects event handlers in our plugin; 
- our plugin receives as a prop a method that FNE defines, that uses a reference to the shortcut we're neighboring so that it knows which shortcut to update when a value is selected. 

The way I see it there are two more obvious ways this can be done: 

1) The Plugin and The Portal both have access to the factory, and we store a reference to the shortcut at the FNE editor level. 
- The Plugin can use a static method on the factory to see if the neighboring shortcut (passed through props) has a completion component and use a props.method to store a reference to that shortcut at the FNE level; 
- our portal will use the factory and a props.reference to the currently-neighbored shortcut to determine which completion component to render, using a props.method to bind a reference to that component to the FNE `this` object; 
- the Plugin uses some props.reference to the completionComponent DOM node passed through FNE to get relevant event handlers; 
- when a value is selected, a FNE function that was passed from FNE -> Portal -> CompletionComponent that can change the data of that shortcut.

2) The Plugin only has access to the factory, and we store references to both the completion component and the shortcut. This approach should be identical except that instead of determing through the factory which completionComponent we should be rendering inside our portal, we just use something that our CompletionPortal receives as a prop and render that as our completionComponent. That prop is a reference to the completionComponent class that we store  at the FNE level. 

If @gregquinn2001  you can let me know which of these is preferable I can continue working on the relevant changes tomorrow AM.

src/notes/FluxNotesEditor.jsx
gregquinn2001: Let's just leave it as is Dylan. What you have works and is not bad. Thanks for entertaining other options. Maybe React 16 can make this more elegant...

gregquinn2001:
OK but I'm approving based on the one part I was looking at - completion component creation and metadata. I'm not doing a complete review.

Dtphelan1:
@mgramigna Finally, it's been rebased.

mgramigna:
Linter error: 

image

Dtphelan1:
pull again buckoo

mgramigna:
We should change the default case's console error for the `completionComponent` getter in `Shortcut.jsx`. When I type #disease status into the editor, I get the following console error:

image

I know that when we have the get help component, that will be the completion component used for ProgressionCreator. But we should probably update the default error message to not mention number subtypes (this was probably just a copy and paste thing).

mgramigna:
Haven't found anything else, and I have confirmed that all resulting Jiras of this PR have been added to the backlog.

👍

Dtphelan1:
Console error has been improved to read better and accuractly describe the current subtype.

PR #672, opened 2019-06-12
SMART on FHIR - reduce console errors and add resourceType configuration

dehall:
Addresses [ASCODCP-1934](https://jira.mitre.org/browse/ASCODCP-1934).

This PR has 2 main impacts:
1) Reduces the number of errors seen in the console when loading patients via SMART on FHIR, by skipping processing any FHIR resources that don't have profiles on them. The profiles are how the `fromFHIR` logic knows what SHR class to instantiate, so if there's no profile on something there's nothing we can do with it.
     - Also adds a fix for one specific SHR class, `shr.base.FindingResult` to let it handle different FHIR types appropriately. (This is a known limitation in the ES6 exporter that also affected the `shr.base.Reason` class) 
2) Adds a data source configuration option `resourceTypes` to specify which resource types should be fetched from the FHIR server. If the config option is not specified it falls back to fetching every resource type the server supports, as defined by the server's conformance statement. I set the `/smartcompass` endpoint to fetch the following FHIR resource types, since based on some quick testing these are the ones the app currently uses: Patient, Condition, Encounter, MedicationOrder, Observation, Organization, Practitioner, Procedure. 

This should be tested on (at least) both the `/smartcompass` and `/synthea` endpoints, since both of those leverage the SMART on FHIR data source. One simple test case is:
 - Launch the Compass app via http://moonshot-dev.mitre.org:4001/?auth_error=&fhir_version_1=r2&fhir_version_2=r2&iss=&launch_ehr=1&launch_url=http%3A%2F%2Flocalhost%3A3000%2Flaunchcompass&patient=&prov_skip_auth=1&provider=&pt_skip_auth=1&public_key=&sb=&sde=&sim_ehr=0&token_lifetime=15&user_pt= -- make sure that FHIR Version is set to R2 (DSTU2) and the launch URL points to your local env /launchcompass
 - Select the default provider and then patient Marshawn McOde
 - Confirm that medications load in the UI
 - Update config.js to remove `MedicationOrder` from the resourceTypes for /smartcompass
 - Rerun the above steps, and confirm that medications no longer appear in the UI

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/dataaccess/utils/fhir-entry-processor.js
+++ b/src/dataaccess/utils/fhir-entry-processor.js
@@ -26,11 +26,16 @@ export default function(responses, patientId, resourceMapper = null) {
     const mappedResources = {};
     const referencesOut = [];
     const allObjects = entries.map(entry => {
+        const resource = entry.resource;
+        // if there's no profile on this resource we can't determine what class to convert it into
+        // so rather than error out just ignore it
+        if (!resource || !resource.meta || !resource.meta.profile || !resource.meta.profile[0]) return null;
+
         try {
-            const result = ObjectFactory.createInstanceFromFHIR(null, entry.resource, entry.resource.resourceType, patientId, entries, mappedResources, referencesOut);
+            const result = ObjectFactory.createInstanceFromFHIR(null, resource, resource.resourceType, patientId, entries, mappedResources, referencesOut);
 
             // shortId here is the standard "resourceType/resourceID" ID format
-            const shortId = `${entry.resource.resourceType}/${entry.resource.id}`;
+            const shortId = `${resource.resourceType}/${resource.id}`;
 
             // this format and the entry fullURL are 2 formats that are used for references
             // so add this object to the map with both keys, so either one could be used for lookups
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "downshift": "^1.31.16",
     "elasticlunr": "https://github.com/mgramigna/elasticlunr.js",
     "es6-shim": "0.35.3",
-    "fhir-mapper": "^1.0.4",
+    "fhir-mapper": "^1.0.6",
     "fhirclient": "^0.1.12",
     "flat": "^4.1.0",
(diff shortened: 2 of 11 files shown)


mgramigna:
Looks good!

The only thing I'm seeing is on `/launchsynthea`, the mapper still has a bunch of errors for unsupported namespace in the `createInstanceFromFhir` function. Looks like this is actually caused by the lack of a full URL in the profiles of the [Synthea Mapper](https://github.com/standardhealth/fhir-mapper/blob/master/src/mapping/mappers/syntheaToV05.js).

We could make this a separate Jira since it seems like this has always been an issue with `fhir-mapper`, but it just became apparent here since the other console errors are gone now.

dehall:
Updated the FHIR Mapper Synthea mCODE 0.5 mapping to ensure everything it adds to profiles is a valid URL, and bumped the reference here to new version 1.0.6

src/dataaccess/utils/fhir-entry-processor.js
```diff
@@ -26,11 +26,16 @@ export default function(responses, patientId, resourceMapper = null) {
     const mappedResources = {};
     const referencesOut = [];
     const allObjects = entries.map(entry => {
+        const resource = entry.resource;
```
rdingwell: Is there some sort of default mapping that can be done?  If FHIR Observation instantiate SHR Observation ...

dehall: We talked about that a little bit some months ago when I initially put this in. I think at the time we decided it was a little duplicative with the FHIR mapper project. I can add something, but I don't think it's really necessary, since a data source "should" either be entirely profiled in which case it's not necessary to do any mapping (ex, via a shim server or a FHIR server containing mCODE'd data) or entirely non-profiled in which case we should configure the full FHIR mapper (ex, via any server that exists today).

PR #674, opened 2019-06-14
fix side effects bug

KeeyanGhoreshi:
Fixes [JIRA 2013](https://jira.mitre.org/browse/ASCODCP-2013), which is that when you switch from the icons view back to the table view, the side effects are missing from the dropdown.  This was because the side effects get loaded when the props change, but not on mount.  Going from icons view to table view remounts the table components, but until the props get changed the side effects list will remain empty.

I originally fixed this by adding the population of the side effects list to the component mount function, but then instead deleted it all from the lifecycle functions and added the list gathering to the header rendering function, since its the only function that really needs a list of all the side effects.  

This fixes the issue and removes the deprecated `componentWillRecieveProps()` function from the component (deprecated in React 16, Flux uses React 15 for now).  

Of course this could be easily redone the original way, just adding the gathering of the side effects to the component mount lifecycle function, if there is some future need to have the list of side effect names in the state of the component.  As far as I can tell, there's no reason to keep that info in the state, since the user will never change it.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/mcode-pilot/components/TreatmentOptionsOutcomesTable/TreatmentOptionsOutcomesTable.js
+++ b/src/mcode-pilot/components/TreatmentOptionsOutcomesTable/TreatmentOptionsOutcomesTable.js
@@ -17,24 +17,16 @@ export default class TreatmentOptionsOutcomesTable extends Component {
         this.state = {
             includedOpen: false,
             comparedOpen: false,
-            sideEffectSelection: 'Most Common',
-            sideEffects: []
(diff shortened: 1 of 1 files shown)


jafeltra:
This looks good to me! The bug is resolved and moving the function to get the side effects to the render doesn't seem to slow anything down, so I think it's good!

PR #675, opened 2019-06-17
Quick fix to prevent null reference in react-minimap

dehall:
Addresses [ASCODCP-2023](https://jira.mitre.org/browse/ASCODCP-2023)

The core issue here is that the react-minimap component is throwing an error when it tries to reference an object that shouldn't be null. Previously (prior to the react-scripts upgrade) this error just printed to console, but now there is an error overlay that pops up and interrupts the user. One alternative or additional change if we want it would be to disable that overlay, but that may require ejecting if our react scripts customizations don't support that specific change. (See https://stackoverflow.com/questions/46589819/disable-error-overlay-in-development-mode )

The exception occurs when the `synchronize` function in `react-minimap.js` tries to call a function on `this.source`. `source` gets set as a "callback ref" ( https://reactjs.org/docs/refs-and-the-dom.html#callback-refs ) and so is guaranteed to be set when everything is rendered synchronously. Where the problem comes in is when the page gets re-rendered asynchronously, ie, when loading a patient via SMART on FHIR. In this case we're actually re-rendering multiple times: the first render is the loading spinner which doesn't include the minimap, the second is the structure of the UI (see screenshot at the bottom of the PR), and the third is the full content of the UI. Before that second render, `source` gets set as expected, but somewhere between that second and third render it gets un-set (same screenshot at the bottom of the PR for the stack trace). Then the synchronize function gets called (via the `setTimeout` in `componentWillMount`)  and errors out. Later before the third render `source` gets set again, and from there the app continues to function as expected. 

I'm sure there is some other way to reorganize or rearchitect either the FN UI or the minimap component to be more async-friendly, but that's going to require more research and time, which I'm not sure this issue really needs.
(You may notice on that react docs link there is a "Caveats with callback refs" -- I tried the change recommended there, but it didn't seem to change anything)

The quick fix implemented here is to just add a null check on `this.source` in `synchronize`. 

This should have no impact on any functionality, just preventing an error from being thrown.

Steps to reproduce:

 - use the SMART launcher at http://moonshot-dev.mitre.org:4001/?auth_error=&fhir_version_1=r2&fhir_version_2=r2&iss=&launch_ehr=1&launch_url=http%3A%2F%2Flocalhost%3A3000%2Flaunchcompass&patient=&prov_skip_auth=1&provider=&pt_skip_auth=1&public_key=&sb=&sde=&sim_ehr=0&token_lifetime=15&user_pt=
 - launch compass at http://localhost:3000/launchcompass
 - select default provider
 - select Minnie McOde

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [ ] Existing tests passed

## Reviewers:

- Try to break the code

Intermediate re-render:
image

### DIFF
--- a/src/lib/react-minimap/react-minimap.js
+++ b/src/lib/react-minimap/react-minimap.js
@@ -262,6 +262,12 @@ export class Minimap extends React.Component {
   synchronize(options) {
       const {width, height} = this.state;
 
+      if (!this.source) {
+        // unexpected condition here, likely caused by re-rendering due to an async data chance
+        // which resulted in the `source` ref getting unset and it has not yet been reset.
+        return;
+      }
+
       const rect = this.source.getBoundingClientRect();
 
       const dims = [ rect.width, rect.height ];


Dtphelan1:
This looks good to me. No errors when I load either Minnie or Marshawn. Comments look good and will serve as a good record for the investigations we've done for this task. Good work!

PR #688, opened 2019-06-25
Shorten names of table cells

ashih2018:

Task 2092 for Longitudinal Lab Visualizer. Added a max width and text-overflow: ellipsis such that any text that will be more than 135 px wide, which is how wide "White blood cell count" is, will be shortened with ... Favoriting functionality was deleted for names as it will be in the starred column which has a PR up right now. In the future, when clicking on names, an action menu will show up. That is a separate task on Jira, this task is just for the shortening of lab/vital names that are too long.
## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/summary/LongitudinalTable.jsx
+++ b/src/summary/LongitudinalTable.jsx
@@ -90,7 +90,7 @@ export default class LongitudinalTable extends Component {
         return (
             <Tooltip
                 placement='right'
-                overlayClassName={`option-tooltip`}
+                overlayClassName={`name-tooltip`}
                 overlay={`${name}`}
                 mouseEnterDelay={0.5}
             >
--- a/src/summary/LongitudinalTable.jsx
+++ b/src/summary/LongitudinalTable.jsx
@@ -17,6 +17,7 @@ export default class LongitudinalTable extends Component {
             }),
             hovered: null,
         };
+        this.refs = React.createRef();
     }
     componentWillReceiveProps(nextProps) {
         if (!_.isEqual(this.props.dataInfo, nextProps.dataInfo)) {
--- a/src/summary/LongitudinalTable.jsx
+++ b/src/summary/LongitudinalTable.jsx
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import ReactDOM from 'react-dom';
 import Table, { TableHead, TableBody, TableCell, TableRow } from 'material-ui/Table';
 import _ from 'lodash';
 import moment from 'moment';
--- a/src/summary/LongitudinalTable.jsx
+++ b/src/summary/LongitudinalTable.jsx
@@ -159,7 +159,7 @@ export default class LongitudinalTable extends Component {
                     <TableCell className={`star-cell star-body ${hoverable}`} onClick={() => { this.toggleFavorites(n); this.props.reorderRows(n.name); }} onMouseOver={() => { this.setState({ hovered: n.id }); }} onMouseLeave={() => { this.setState({ hovered: null }); this.renderStar(n.name, n.id); }}>
                         {this.renderStar(n.name, n.id)}
                     </TableCell>
-                    <TableCell>
+                    <TableCell className='name'>
                         {n.name}
                     </TableCell>
                     <TableCell>{n.unit}</TableCell>
--- a/src/summary/LongitudinalTable.jsx
+++ b/src/summary/LongitudinalTable.jsx
@@ -86,7 +86,20 @@ export default class LongitudinalTable extends Component {
         }
         return <div />;
     }
-
+    renderNameCell(name) {
+        return (
+            <Tooltip
+                placement='right'
+                overlayClassName={`option-tooltip`}
+                overlay={`${name}`}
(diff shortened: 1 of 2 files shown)


sgober:
[SO1] Almost all the css properties in `LongitudinalTable.scss` are designated with an `!important`.  Is there a way to refactor the html/scss to avoid doing this?  Ideally we should only use `!important` in specialized cases.

mgramigna:
@sgober Using `!important` isn't a fault of anything here. The reason that it comes up so much is that we are not overriding Material-UI components' CSS classes properly. There is a defined API in Material-UI for overriding styles on their components that we are not using.

As such, the `!important` flag is used when we need to override a specific style that is present on that Material-UI component. Ideally we should be using `makeStyles` which is talked about in the Material-UI docs. Unless you see an obvious way to refactor this stuff while still using Material-UI's table, I'm not sure there's much to be done on this end if we want to continue using scss the way we are across the whole project.

mgramigna:
However if there are `!important`s here that are not actually overriding anything and are there just because, I agree that we should remove those. But for the most part any of them that touch a property that Material-UI's CSS also touches, we might be out of luck.

sgober:
@mgramigna I agree, I definitely think some of them could be removed.  Might take some investigating though

dreeves4321:
@mgramigna, on my next pass on styling, I'm hoping to tackle a few of these MUI components that have hard to correct styles. Have you used makeStyles or withStyles? I'm a little confused by them, and how to properly use them. if you have any experience there, I may request your help getting started. Thanks!
Dan

mgramigna:
@dreeves4321 I do a little bit. I don't think `makeStyles` is even a thing in our current version of material anyway, but `withStyles` definitely is. I would love to help out whenever! @Dtphelan1 might have some useful insight as well :)

sgober:
@ashih2018 
[SO1] I just spoke with @mgramigna and we agree we should investigate which css properties need the `!important` and remove the ones that don't.  After that we can punt the refactor to override the material classes to another time.

[SO2] Another thought, since we are using sass here we should try and nest the `.scss` file.  We do this in other `.scss` files if you need an example (try `FluxNotesEditor.scss`

Also this now has conflicts with master so it will need a rebase.

ashih2018:
@sgober I just rebased the branch and I removed unused `!important`

sgober:
This looks good @ashih2018!!

One other thought/question... do we think it makes sense to have a tooltip on that cell with the full name so the user can see what is being cut off on hover?

ashih2018:
@sgober i was thinking that would maybe be an action for the action menu when you click on a name, it would show the full name and other actions maybe?

Dtphelan1:
@dreeves4321 do you have any thoughts on where the tooltip would fit in here?

dreeves4321:
Two thoughts:
1) If this is possible, I think it would be great. With no delay, expand over the table on hover.:
image

2) Otherwise, make the ellipsis hoverable. 
image

sgober:
@dreeves4321 I think #2 would be appropriate for this PR and then we can enhance it later

ashih2018:
@sgober I talked about it with matt and dylan and we agreed that it would be difficult to get a hover on the ellipsis because the ellipsis is css not an element or component

dreeves4321:
Okay, then scrap the styling on the ellipsis and just hover on the whole word. But make a short delay on the hover to avoid unwanted popups (which I was hoping to avoid by placing the hover over a smaller target). There shouldn't be too many labs much longer than the limit.

sgober:
[SO4] I am seeing weird placement of the tooltip if the name is long enough to force the ellipsis
 
Screen Shot 2019-07-01 at 9 15 47 AM

ashih2018:
[SO4] this is because the tooltip is placed right after the word, so if the word is longer it will be placed right after it. I tried placing it on top, bottom, and left, but they all looks worse than putting it right. Furthermore, me and matt discussed that fixing this probably wouldn't be worth the time because there is no easy fix.

sgober:
@Dtphelan1 can we just force it to be at the end of the name cell instead of putting it at the end of the word?  I feel like that would look better and not be super complicated

ashih2018:
@sgober fixed! this pr should be ready to be merged unless if you spot anything else that should be changed!

src/summary/LongitudinalTable.jsx
```diff
@@ -87,6 +86,20 @@ export default class LongitudinalTable extends Component {
         }
         return <div />;
     }
+    renderNameCell(name) {
+        return (
+            <Tooltip
+                placement='right'
+                overlayClassName={`option-tooltip`}
```
Dtphelan1: Did we just copy this class over from another file? We should come up with our own class here so we can intentionally style the width of the cells so they're not extremely large when unnecessary. IMO.

Dtphelan1:
Last comment - is the !important z-index required?

Dtphelan1:
Looking at this now!

PR #691, opened 2019-06-28
Allow multiple insertions of #reason

mgramigna:

For ASCODCP-1936. Originally scoped as writing a completion component for multi-selection, but shifted towards allowing #reason to be inserted multiple times using the same completion component as single-choice shortcuts.

Can now enter #reason to select a value. Can enter as many #reason shortcuts and select as you want until you have selected all choices in the value set.

**NOTE**: While doing this task I uncovered something on `master`/fnorg where the value set options for #reasons are always in the suggestion portal regardless of if they are in the editor or not. Created ASCODCP-2104 as a result.

Potential thing to keep in mind for the future: With this approach, I think clinician customization of the expanded text will need to support adding multiple incomplete #reason labels since we need to support 0 .. many. 

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/shortcuts/CreatorChild.jsx
+++ b/src/shortcuts/CreatorChild.jsx
@@ -15,6 +15,7 @@ export default class CreatorChild extends Shortcut {
 
     initialize(contextManager, trigger, updatePatient = true) {
         super.initialize(contextManager, trigger, updatePatient);
+        super.determineParentContext(contextManager, this.metadata["knownParentContexts"], this.metadata["parentAttribute"]);
         let text = this.determineText(contextManager);
         if (!Lang.isUndefined(text)) {
             if (Lang.isArray(text) || text === 'date-id') {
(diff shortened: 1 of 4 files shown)


jafeltra:
This looks good! I like this approach and I personally don't find adding multiple #reason's too confusing. Two things I noticed that may or may not be in scope of this PR:

[JA1] Add "#reason (uncompleted) some other text #pathology (added in however you want)" now if you go back to #reason to complete it, Pathology is still in the portal of suggestions, even though it is already in the editor. Complete to pathology and it can be set twice, but only shows up once in the TDP.
[JA2] After completing the steps in JA1, if you go to the end of the line and delete the second "Pathology," the TDP shows no data for the rationale, even though there is a pathology added to the note.

sgober:
[SO1] I am getting an error when I type out what is in the screenshot below, then move the cursor up next to the condition and then back down to the end of the last reason

Screen Shot 2019-07-01 at 9 07 08 AM

Screen Shot 2019-07-01 at 9 07 15 AM

mgramigna:
@sgober SO1 is already a Jira that we found in #665

sgober:
@mgramigna my b, too many JIRAs for me to keep track of

mgramigna:
That Jira that I mentioned that I found in the PR could also encompass JA2: https://jira.mitre.org/browse/ASCODCP-2104 since that is the root reason that we can type multiple pathologies in the first place.

Looking into JA1 now

mgramigna:
*TL;DR: JA1 = Jira?*

---

So for JA1. This happens because the value selection options are only set on initialize. When you insert #reason, the valueSet that it uses is determined on initialize. When you then insert #pathology, it never updates the valueSet from #reason since it doesn't need to get re-initialized.

Simplest way would be to have the valueSet re-compute whenever the completion portal opens? There could be a more at-large refactor here for the shortcut model to make this work nicely. Either way, I'm gonna go on a limb and say it shouldn't be resolved for this PR. Thoughts?

mgramigna:
Also, JA1 might not even be a use case in the MVP in my opinion

jafeltra:
I agree. The root cause of it seems separate from this task so I think we're okay to not include it in this PR and make a Jira.

PR #698, opened 2019-07-08
References Cleanup

Dtphelan1:
Pretty straightforward - ran the backend tests and did a cursory glance for functionality's sake. But removed the PR Template since there's little to look at.


danlee1025:
Not sure if we still even use the `stagingSheet.pdf` but we may want to remove the AJCC reference there too.

Dtphelan1:
Good idea - I'll do that shortly.

Dtphelan1:
I'm going to be talking with May Terry briefly as well to make sure this falls in line with the best practices established by the Modeling team.

rdingwell:
I see that things have been changed from AJCC to GIST.  Is GIST appropriate for Breast Cancer Staging?  GIST stands for gastrointestinal stromal tumors correct?

src/lib/cql-execution/example/cql/PALLASEligibility.cql
```diff
@@ -35,7 +35,7 @@ concept "Progesterone Receptor": { "Progesterone receptor [Interpretation] in Ti
 concept "Hemoglobin": { "Hemoglobin A1c [Mass/volume] in Blood" } display 'Hemoglobin A1c'
 concept "Platelets": { "Platelets [#/volume] in Blood" }  display 'Platelets [#/volume] in Blood'
 concept "HER2": { "HER2 panel - Tissue by FISH" } display 'HER2 panel - Tissue by FISH'
-concept "AJCC": { "AJCC cancer staging manual [PhenX]" } display 'AJCC cancer staging manual'
+concept "Staging": { "AJCC cancer staging manual [PhenX]" } display 'Cancer staging manual'
```
rdingwell: Do we want this reference to AJCC Removed?

dehall: That's the canonical display text for a specific LOINC code. I had suggested offline that in that case it made more sense to leave the canonical text rather than change it

Dtphelan1: Exactly right. Since all we're doing is referencing a LOINC code, and since LOINC's licensing is generous, this is a problem b/t LOINC and AJCC, not us 👍

Dtphelan1:
After conversations with May, Greg, Andre and Zeshan the short-term strategy is to remove any references _we created_ to AJCC Staging systems or Staging computation values. This includes the automatic computation of Stage (which is just an encoding of AJCC's algorithm). Consequently we might need to enable the manual provisioning of a staging value, instead of the automatic computation of the value at the shortcut level. 

Long-term, we are going to need to enhance our CQL and Clinical Trial Matching to use valuesets, instead of manually specifying that they use AJCC codesystems. These valuesets can then provisionally reference the AJCC codesystem - or not - based on the licensing restrictions of the deployment location (e.g. BHW, BCH, COZE, etc...) I'll make a JIRA to address this.

Dtphelan1:
Actually, on the tails of additional conversations with Andre and Greg, this is sufficient for what we're trying to do now. Places we reference AJCC related codes might fall under ta fair-use clause, and other places we use AJCC-related information we were able to retrieve from other external locations (e.g. LOINC, SNOMED, etc...). The thinking, as stated above, is that AJCC should have grief with them and not us so much. If they do have issues with the ways we've used their IP under what we thought might fall into categories of Fair use, they can let us know and we'd be happy to comply in a timely manner. 

tl;dr - We're keeping the changes we've made, not adding more at this time. Wil make a JIRA to track this moving forward.

Dtphelan1:
Travis is happy, I'm gonna merge 

Thanks Sarah!

PR #708, opened 2019-07-12
Next patient

ashih2018:

HOW TO RUN: Go to flux wiki and go to pre release testing script. On terminal, run the scripts for ella and ihanos. On line 1628, change the time from 16:00 to 12:00 and the next patient button should work.

Created a next patient button for JIRA task #1985. When clicking on the next patient, the patient with the closest expected performance time (appointment time) to the current patient's appointment time is loaded and rendered onto the page. 
I also did not know where to put the Next Patient Button, the new component I created, (folder-wise), so I put it in the same folder as where it is being used, POCDashboard. Is this the right place to put this component?
Possible shortcomings:

- When the button is clicked, the next patient is rendered correctly, but the following error is thrown: 'index.js:1446 Error:  TypeError: Cannot read property 'isDemo' of undefined.' This occurs in the TreatmentOptionsSection as section.data[0].data_cache is a promise that is undefined at the time. This object is passed in from the scatterplot visualizer, so this should be checked out in the near future.
- The current functionality assumes that each patient only has one appointment a day, as we do not currently associate a page with an appointment time (we associate a page with rendered data with the patient, so if there are two appointments for a patient, Flux Notes does not know which appointment the data is for and automatically assumes it is for the earliest appointment that day.
- Button outline does not have a specific color provided by Dan Reeves or in the styles/variables.scss file, so I matched it up as close as it looked like on the markup.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


mgramigna:
[MG1] Looks like we have some extra whitespace at the bottom of the app now, along with a rightmost scrollbar (tried to get a screenshot showing this). I'm assuming this is due to some CSS related to the new next patient button (which looks great btw :)). Let's try to remove this whitespace and scrollbar

image

mgramigna:
[MG2] Add `src/dashboard/NextPatientButton.css` to `.gitignore`

src/containers/PointOfCareApp.jsx
```diff
@@ -314,8 +318,10 @@ export class PointOfCareApp extends Component {
                                         dataAccess={this.dataAccess}
                                         highlightedSearchSuggestion={this.state.highlightedSearchSuggestion}
                                         loginUser={this.state.loginUser}
+                                        loadPatient={this.loadPatient}
                                         patient={this.state.patient}
                                         preferenceManager={this.preferenceManager}
+                                        test={this.test}
```
mgramigna: [MG3] Assuming we don't need this prop anymore since it was just for testing??

ashih2018:
MG2 AND MG3 resolved!

ashih2018:
MG1 resolved!

src/dashboard/PointOfCareDashboard.jsx
```diff
@@ -145,15 +157,59 @@ export default class PointOfCareDashboard extends Component {
         return this.state.showPOC ? this._unselectedTdpIcon() : this._selectedTdpIcon();
     }
 
+    renderNextPatient = () => {
+        const dataAccess = new DataAccess('HardCodedMcodeV01DataSource');
```
jafeltra: [JA1] Should the DataAccess we use here be 'HardcodedTabletMcodeV01DataSource'?

dreeves4321:
Looks very nice! Two bugs and a 1 maybe:
[DR1] For any size other than 1366x1024 (at least in the simulator) the button doesn't situate properly in the bottom left. For example, other ipads are slightly different.

[DR2] The button shifts by 1 px in both horizontal and vertical when switching between record and poc modes.

[DR3] There is a hovered state on the buttons. They shouldn't ever get called on the ipad, but in the simulator they seem to upon clicking. Not sure if that's a bug, or just let it go.

src/dashboard/PointOfCareDashboard.jsx
mgramigna: Theoretically yes, but we actually should not be constructing a new DataAccess instance here. The dashboard takes dataAccess as a prop. Instead of constructing a new one, use `this.props.dataAccess`

ashih2018: JA1 Resolved!

ashih2018:
[DR1] and [DR2] resolved! For [DR3], Julia said we don't have to solve this for now, as there does not look like an easy solution to this problem, and it is not a big issue for this PR.

mgramigna:
[MG3] I noticed that after switching to Ella, the Point of Care view doesn't update with her note correctly. Ella's note should only have a disease status placeholder, whereas Ihanos should have both a diseases status placeholder and a toxicity placeholder.

Did some (surface-level) digging, and it appears that `this.props.patient` on `PointOfCareDashboard` never gets updated to reflect the new patient that you have switched to. That is the root cause of the note content not showing up properly.

mgramigna:
Actually, upon looking a little more, it looks like the logic for setting up the note with correct placeholders happens in the constructor, so this will not get re-called when you load a new patient.

We might want to just create a Jira for this, but I'm gonna confirm with others before committing to that.

PR #713, opened 2019-07-19
Insert templates directly

jafeltra:

This PR allows you to insert templates directly into the editor with incomplete shortcuts, eliminating the need for picklists. Individual shortcuts inserted via the context tray that used to use picklists also insert directly as incomplete items.

Because picklists are no longer used and templates are inserted directly, I was able to take out a lot of unused code. The main types of code I removed was picklist option components, cases to handle when note assistant was in picklist mode, and functions used to cancel template insertion and reset the editor. I also noticed that TemplateViewModeContent was never used and seemed to be replaced by TemplateSelectionView, so I removed that component and it's child component (TemplateForm). I also fixed a test that use the picklist and added a new test to ensure incomplete items are inserted directly without a picklist.

Some potential shortcomings:
- When you insert the follow up template and select a value for @condition in the editor through the Completion Component, @oncohist no longer automatically fills in. I made a function that has the old functionality for updating child shortcuts after the parent has a value and put a commented out call to the function after a completion component value is selected. Right now, uncommenting the line will not update the child component (@oncohist) when the parent gets completed (@condition) because when we insert incomplete shortcuts, we remove them from context. Therefore, when the incomplete child is added to the editor, it doesn't get added to the parent because it can't determine its parent context since it isn't in context. Removing the line to remove an incomplete shortcut from context (FluxNotesEditor.jsx ~line 342) and adding the call to updateChildShortcuts causes @oncohist to update. I figured enabling this might be beyond the scope of this task, but I can re-evaluate if reviewers think this should be addressed now. Otherwise, I can add these notes to [Jira 2026](https://jira.mitre.org/browse/ASCODCP-2026).

- The UI test files still reference picklist classes. We never run those tests and they will now fail if we do, but it wouldn't be simple to just remove the references and get the test to pass, so I left them. I could just remove all those tests if we want.

- The pre-release script will need to get updated but since this isn't getting merged to master, I haven't done that yet.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


dreeves4321:
slick. not sure if this is part of this task, but it would be nice if incomplete items could open the drop menu with a single click. i need a double click for anything to happen.

karenzheung:
KZ1 I get an error when I type #to when doing #reduce medication
 
Screen Shot 2019-07-22 at 11 06 16 AM

jafeltra:
Thanks @dreeves4321! Are you clicking/double clicking on the actual incomplete item? I don't think we have much (or any) specific handling for clicking on incomplete items themselves. Clicking at the end of the incomplete item should open the menu after one click. I can add a future task for open the menu when clicking directly on the incomplete item, but I think it would fall out of the scope of this PR.

dreeves4321:
it looks like double clicking works because it highlights the whole word and moves the cursor to the end of the word. From a UI standpoint, you want a single click to open the menu. So it would be good to add a task to either 
A) move cursor to end of word when tapping on an incomplete capture shortcut
or
B) open menu directly when tapping on an incomplete capture shortcut, regardless of cursor position.
Not sure which is better and I have no idea which is more doable!

karenzheung:
When I insert the follow-up template and select GIST for condition @ONCOHIST doesn't update like you said. But I can arrow key into it and select oncohist from the suggestion portal which will insert it. But when it inserts oncohist it replaces only the part of the shortcut that was to the left my cursor and the text on the right remains as incomplete so you can't sign the note unless you delete it. 

Screen Shot 2019-07-22 at 11 38 24 AM

I'm not sure how much of an issue this is since getting the child shortcut to update would prevent this.

karenzheung:
When I insert the follow-up template and select GIST for condition @ONCOHIST doesn't update like you said. But I can arrow key into it and select oncohist from the suggestion portal which will insert it. But when it inserts oncohist it replaces only the part of the shortcut that was to the left my cursor and the text on the right remains as incomplete so you can't sign the note unless you delete it. 

Screen Shot 2019-07-22 at 11 38 24 AM

I'm not sure how much of an issue this is since getting the child shortcut to update would prevent this.

jafeltra:
Thanks Karen! I'll look into KZ1.

For the oncohist issue, I noticed that @oncohist is now editable when it's incomplete, which I think has to do with the fact that it is editable once you insert it. I actually think it is pretty easy to add a check for `isComplete` along with `isEditable`, which we do check before we allow editing shortcuts. I can add that in here.

jafeltra:
Thanks @dreeves4321! I'm not exactly sure off the top of my head which was easier. We were actually talking about how it might be possible and not too difficult to do your option A, but I can add a task to open the portal when clicking on the incomplete and we can decide on the implementation when we do that task.

jafeltra:
I looked into KZ1 and it appears to be breaking on `shortcut-entry-overhaul` too, but you fixed it in your `expand-text` PR, so I think it's okay to leave it as is here and get the fix in from your PR. The fix is in ShortcutManager.js around line 344. You updated the way we calculate the numberOfValidTriggers.

mgramigna:
For situational awareness: I am taking over @sgober's review of this PR since she is pre-occupied with testing `master`/other PRs

mgramigna:
This is awesome! I do think that the ONCOHIST stuff requires some investigation. But other than that, I haven't found anything else.

karenzheung:
Aside from the ONCOHIST issue everything else looks great

jafeltra:
After some investigation into the @oncohist issue and talking with @Dtphelan1, we are going to hold off on addressing updating @oncohist when an @condition is selected. Of the options I explored, keeping incomplete shortcuts in context was the simplest solution, but we felt that there were too many drawbacks since it would put child shortcuts in context and mess up breadcrumbs. Other options I looked at briefly were allowing child shortcuts to add to inactive contexts, but that breaks the idea of local contexts, and adding child shortcuts outside of the initialize function, but that seems to take on too much of the work that should be done in [Jira 2026](https://jira.mitre.org/browse/ASCODCP-2026). The @oncohist updating should be resolved as part of that Jira, and we are planning to do that task as part of this shortcut guidance MVP.

I've also rebased the branch. Given that we are holding on the oncohist issue, this PR should be ready for re-review.

PR #722, opened 2019-07-30
Basic throttling of onScroll

Dtphelan1:
Pretty straightforward change coming out of the IE investigation - increases the throttle on the minimap, and uses the throttled function as the onScroll handler. Results in remarkable performance improvements on IE, at the cost of the actual minimap grey-box looking a bit laggy. That's a tradeoff I'm comfortable with.

### DIFF
--- a/src/lib/react-minimap/react-minimap.js
+++ b/src/lib/react-minimap/react-minimap.js
@@ -38,7 +38,7 @@ export class Minimap extends React.Component {
       this.init = this.init.bind(this);
       this.up = this.up.bind(this);
 
-      this.resize = _.throttle(this.synchronize, 100);
+      this.resize = _.throttle(this.synchronize, 200);
 
       this.state = {
           children: null,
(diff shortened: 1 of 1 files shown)


gregquinn2001:
This is awesome @Dtphelan1 !!!! I just tried it on IE and it almost makes me willing to actually use IE! Nice work.  The only thing I noticed (may not be related to your changes - I didn't check master sorry).
[GQ1] Switching the visualization for a section in TDP doesn't update minimap until I scroll the targeted data panel. I seem to remember we had a jira about the minimap automatically rescaling sections when the visualization changes (and therefore the height of the section changed) so maybe disregard!

Dtphelan1:
Great catch and that is totally captured by another JIRA and shows up on Master! 👍

jafeltra:
This looks good. I think there is a tradeoff here where the scrolling looks a little jumpier on Chrome, but IE is definitely much more usable and I think the tradeoff here is worth it.

PR #725, opened 2019-07-31
adding mapper to mcode data source 

rdingwell:
Very simple PR to add the FHIR mapper into the mCODE data source.  This allows for patients that do not have mCODE profiled resources to be loaded into the application without it crashing.  The mapper currently just goes through all resources from the server and adds appropriate shr profiles to them so the underlying es6 class will import them correctly in the fromFHIR calls. 

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [ ] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/public/config.js
+++ b/public/config.js
@@ -32,7 +32,8 @@ CONFIG = {
         isExact: true,
         dataSource: 'McodeV05SmartOnFhirDataSource',
         dataSourceProps: {
-            resourceTypes: ['Patient', 'Condition', 'Encounter', 'MedicationOrder', 'Observation', 'Organization', 'Practitioner', 'Procedure']
+            resourceTypes: ['Patient', 'Condition', 'Encounter', 'MedicationOrder', 'Observation', 'Organization', 'Practitioner', 'Procedure'],
+            mapper: 'syntheaToV05'
         },
         logoObject: {
             path: './compass-logo.png',
--- a/public/config.js
+++ b/public/config.js
@@ -30,7 +30,7 @@ CONFIG = {
         display: 'Compass™',
         app: "CompassApp",
         isExact: true,
-        dataSource: 'McodeV05SmartOnFhirDataSource',
+        dataSource: 'GenericSmartOnFhirDstu2DataSource',
         dataSourceProps: {
             resourceTypes: ['Patient', 'Condition', 'Encounter', 'MedicationOrder', 'Observation', 'Organization', 'Practitioner', 'Procedure'],
             mapper: 'syntheaToV05'
--- a/src/dataaccess/McodeV05SmartOnFhirDataSource.jsx
+++ b/src/dataaccess/McodeV05SmartOnFhirDataSource.jsx
@@ -1,6 +1,6 @@
 import IDataSource from './IDataSource';
 import processFHIRResources from './utils/fhir-entry-processor';
-
+import mappers from 'fhir-mapper';
 import 'fhirclient';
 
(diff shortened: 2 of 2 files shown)


mgramigna:
Per discussion, +1 to what Dylan said. I think there's still some work that can be done to improve how this whole data source/mapping to mCODE works, but given what we have I think it makes sense to change the DataSrouce of `/smartcompass` to be the Generic DSTU2 SoF one.

The assumption of the `McodeV05SmartOnFhirDataSource` is that it takes mCODE v0.5 as input and outputs FN Object Model. For now, I don't think we should break that assumption

gregquinn2001:
+1 as well! That was the original idea of the 2 different data sources!

rdingwell:
That is a possibility.   I did not work on either of those so was unclear as to the differences and why there was a mapper in one and not the other

Dtphelan1:
I've made the changes to use the other DataSource @rdingwell I hope that's alright. Going to review as best as I can now that that's changes.

Dtphelan1:
This looks good to me now.

dehall:
So this loads patients and seems to apply the profiles, but for some reason `/smartcompass` no longer uses the cancer condition template, it only uses the basic template. Ex, on master, load Marshawn McOde05 from moonshot-dev, the condition "Malignant tumor of colon" displays with the cancer template correctly via `/launchsynthea` and `/launchcompass`. Now on this branch, it still displays correctly via `/launchsynthea` but via `/launchcompass` it only shows the basic condition template. Not sure why that would have changed.

dehall:
Another thought: since we know that the `mapper` is going to have to change in every deployed instance of Compass anyway, is this really necessary? I.e., the syntheaToV05 mapper is never going to be the actual mapper used, so rather than fiddle with this now and then change it again later, why not just wait until later and set the right mappers when we have them?
(That said I'm still curious about why the templates aren't working as expected so I'd still look into that)

rdingwell:
I actually question the need for there to be 2 distinct data sources as the generic one is derived from mcodev05 one.   The only difference in functionality is the inclusion of a mapping ability.  The mCODE05 functionality as is could then be had by just not including a mapper in the generic class.  I would suggest that we merge this in to a single datasource.

Dtphelan1:
Do we want to do that as a part of this PR? Also, I think the problems with regular conditions being viewed as cancer conditions in terms of the sections shown is a problem on `master` not just this branch. If true, we should make a separate ticket to investigate.

Dtphelan1:
Alright, here's the scoop: 

Regarding the difference in sections, Dylan Hall, Matt and I determined this morning that the synthea mapper doesn't seem to be mapping cancers to FluxCancerDisorderPresent properly. This bug wasn't noticed on the /launchSynthea endpoint because that endpoint uses metadata on every condition that looks like the cancer-specific metadata for /launchcompass. I'll make a task to fix the mapper for cancers appropriately. 

Regarding the two dataSources, there's a discussion to be had that maybe we can have today in lieu of standup (since that's been cancelled)? @rdingwell let us know if that works for you.

Dtphelan1:
Feels like this PR is being held up by the Datasource discussion still - @rdingwell or @dehall do you have additional thoughts?

Dtphelan1:
Closing this for now. We can address questions about the generic vs. specific inheritence ordering and update the mapper to the v09 mapper when we have one.

PR #736, opened 2019-08-14
Set default font-size - fixing IE font-size

Dtphelan1:
Fixes our default font-size to 16px, overriding the browser-specified defaults. This makes sure fonts look the same across browsers. Please test on IE and any other browsers you may have for cross-compatibility. 👍

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -6,6 +6,10 @@ body {
   color: #232323;
 }
 
+html {
+    font-size: 16px;
+}
+
 ::-webkit-scrollbar {
     width: 10px;
     height: 12px;


Dtphelan1:
Rebased with `master`

karenzheung:
On IE, in the timeline when I click on 6M or some other timeperiod the name of the medication are no longer in view and if I hover over the blue medication event the name of the medication goes off screen and gets cut off

Dtphelan1:
Good observation Karen. I can make a ticket for that, as it falls outside the scope of this PR. We'll make sure to start working on that ASAP since the Timeline is one of our Pilot scenarios - good find!

jafeltra:
This looks good to me! The font sizes match across Chrome, IE, Safari, and Firefox. I did notice that the font itself on Firefox looks different. Is that even something we care about?

Dtphelan1:
I'm not too worried about it now. I'd ignore it for now until it becomes something that actively perturbs our users.

jafeltra:
You beat me to the rebase!

PR #739, opened 2019-08-14
Fixed indexing of medications in MedicationsColumnsSections.jsx

Dtphelan1:
Another painfully straightforward PR. Looks like the type that we were using in the MedicationsColumnsSection wasn't able to handle data formatted the way that medications are. On the one hand, it's troubling that the format of the data was so different that the application crashed. On the other hand, I don't think it's a priority to address that schematic difference right now.I'll create a JIRA to address that at a later point.

Testing can be performed by changing all of the Medications on Ella to reference her Fracture instead of her Cancer. The DefaultMetadata, used by all non-cancer conditions (like Fracture), calls on the MedicationsColumnsSection, which by default uses the tabular format for rendering the Medications Visualizer. In addition to confirming the functionality of this usecase (which would have been broken before), please test against other patients as capable!

### DIFF
--- a/src/summary/VisualizerManager.jsx
+++ b/src/summary/VisualizerManager.jsx
@@ -242,6 +242,7 @@ export default class VisualizerManager {
     }
 
     getVisualizer(dataType, visualizerType) {
+        // TODO: Use a `find` or `pluck` here to get the element by reference, instead of getting an array and chosing the first thing
         let result = this.visualizers.filter((viz) => {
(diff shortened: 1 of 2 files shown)


Dtphelan1:
Great question Jules - looks like I didn't make that task when I thought I did. I think our conversation about a new JIRA this morning should be sufficient, but I wanted to make sure we had a public record of me recognizing your good comment!

PR #741, opened 2019-08-15
External fhir mapping

rdingwell:
This PR bumps the version of and externalizes the fhir-mapper so that it can be used within the flux code base as well as be overridden from outside of the codebase.  A webpack config override is added to declare the library as an external as well as a plugin to copy the mapper distribution file to the build directory as needed.  The distribution file is included in the index.html file as a script tag and makes the mapper available via a global window variable called Mapper. 

In addition to the externalization of the mapper there are a few fixes that are included to deal with the Cerner sandbox environment. First is the explosion of the smart on fhir scopes within the launchcompass route.  Cerner does not support wildcard routes so the individual routes need to be declared.  This should have no effect on other fhir servers and works against the smart on fhir community launcher.  

A separate mapper file is included that has addtional codes for mapping sandbox TNM values , to a certain extent, as well as additional cancer types.  This mapper defaults to the syntheaV05 mapper.  

Within FluxCancerConditionPresent there is a patch for dealing with findingTopicCodes that do not have a coding.   Null pointers errors were being thrown from the getters that tried to make use of data that did not exist. 

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [ ] Existing tests passed

## Reviewers:

- Try to break the code


dehall:
Just to confirm my understanding, is the idea that if we want to add or a mapper after building, we would manually rebuild the fhir-mapper and replace the `static/js/fhir-mapper/app.bundle.js` file?

rdingwell:
@dehall not quite, but that is an interesting idea as well.   The idea here was more along the lines of the discussion in slack about a mapper that could be included in the config file or a separate js file that would then be passed through the config for the data source.  If you look at GenericSmatOnFhirDstu2DataSource you'll see that it accepts either a string or an object as the mapper now.

dehall:
Ok so if we do want to add a new mapper via that route then it will either have to be defined completely within `config.js` or we'll have to modify the `public/index.html` file to add a reference to the new script file that it lives in.  Would it make sense to add an "empty" mapper file to start with and reference it in index.html? That way we only have to update index.html once and replace just that one mapper each time.

Dtphelan1:
Looks like @dehall still has an outstanding question here that should be addressed. I'm sufficiently satisfied with this approach for now. Talked with Rob to help me get a better understanding of what's happening and I think it makes sense; but this general type of webpack-level configuration just feels pretty funky, understanding or not...

rdingwell:
@dehall @Dtphelan1 I agree with the idea of a placeholder for external mappers and added that to the index.html file.  I also added an empty mapper.js file as the point of entry for defining the external mappers and rebased the branch on master.

dehall:
I tested this out on IE, fearing that the blank JS file might cause issues, and I don't think the blank file is what's causing the issue but suddenly things don't seem to work on IE at all. I'm just getting a blank page with no console error. Looking further into it now.

rdingwell:
@dehall does the update that you pushed require the updated fhir-mapper to work correctly? Does it need to be published?

dehall:
Yes/yes

PR #761, opened 2019-09-03
Initial implementation of disabled expand

Dtphelan1:
No mockups exist for what the disabled-when-complete getHelp menu should look like, so this is my initial attempt. People should let me know what they think!

### DIFF
--- a/src/context/ContextGetHelp.jsx
+++ b/src/context/ContextGetHelp.jsx
@@ -133,11 +133,28 @@ class ContextGetHelp extends React.Component {
         );
     }
 
+    renderIsCompleteMessage() {
+        const initiatingTrigger = this.props.shortcut.getDisplayText();
+        return (
+            <ul className="context-get-help" ref="contextGetHelp">
+                <li
+                    className="context-get-help-li"
+                >
+                    <span className="context-get-help-text">
+                        <i>{initiatingTrigger} is already complete</i>
+                    </span>
+                </li>
+            </ul>
+        );
+    }
+
     render() {
+        // If the shortcut we're resposible for is complete, display a message to the user to avoid confusion
+        if (this.props.shortcut.isComplete) return this.renderIsCompleteMessage();
+        // Else we should display all our getHelp message
         const initiatingTrigger = this.props.shortcut.getDisplayText();
         let iconClass = 'fa fa-angle-';
         this.state.selectedIndex === -1 ? iconClass += 'down' : iconClass += 'up';
-
         return (
             <ul className="context-get-help" ref="contextGetHelp">
                 <li


sgober:
@Dtphelan1 what are your thoughts on disabling the expand menu option (grey out and don't allow clicking) so that the get help menu looks the same to the user?  I was under the impression that going forward there could potentially be more get help options in the menu so enabling/disabling them while keeping them in the list seems like the right approach if that is the case.

We should also get feedback from @dreeves4321 and/or @edwinychoi :)

mgramigna:
@sgober we briefly discussed that and thought it might be weird to have a menu with only one option that is not clickable. Thinking in the context of the MVP, where there is not going to be an open GUI option, I think that was part of the motivation for this. In the future where there is an open GUI option, I think the approach you mentioned might make more sense to pivot to in the long run, but it seems strange when expand is the only option in the menu for the MVP. 

That's just my two cents

sgober:
@mgramigna I can totally that and pros and cons to both sides.  For me it seems weird to not reference the get help menu even if expand is completed.  I guess it just seems like it could be confusing because it's not consistent with what it looks like when it's not expanded.

Idk, there's not really a good answer because it seems over the top to do what I suggested but at the same time it's consistent with what we've already done.  I think I still vote for consistency though

Dtphelan1:
Long term: Absolutely we'll need to figure out what the UI looks like to communicate that we have a list of options, and some of them are disabled, but not all of them. 

Short Term, Arg 1) We don't have mockups for what that case actually looks like, and we definitely don't have a detailed understanding of all the user interactions involved in the case where we have a disabled menu option (How is it styled? Does it have a different bg-color? What happens when you hover it? How do you communicate that selecting it will do nothing? That it did nothing?) 

Short Term, Arg 2) We aren't actually helping uesers when there's only one menu item, that they need to interact with a menu to see, and only find out after that interaction that it's disabled. My concern is this might instill frustration with the user. From their perspective, they aren't thinking about "The Get Help Menu" (at least I don't think so). From the user perspective, we're establishing that when you move your cursor next to a shortcut it will open "Some Component" - sometimes a completion component, a date picker, a help menu, and now there's a case (for the short term) where the thing that opens is a display that says there's nothing we can help with. 

That was my thinking going into this. Not only is it quicker to implement the functionality this way since we don't actually have a clear understanding of what the user interactions look like in the case of a disabled menu option, but I also think it's less frustrating from a user perspective to see that there's no help we can give them rather than a help menu that has no helpful options.

edwinychoi:
I agree with @Dtphelan1 that they're probably not thinking about the "get help menu". Is one approach to show the options right away, without having to expand a "help menu"? Showing the options themselves would then serve as initial help up front for the user.

I feel that the time the users may need/want help the most is when they're just starting a new #shortcut. If that's true, then it also reduces the priority of showing a help menu later on when they're returning to previous #shortcuts. If they choose to not select from the initial options in the "help menu" and go off continuing to type what they want, then when they return to a previous shortcut where there's nothing we can help with, is one option to not show any display that communicates "there's nothing we can help with"? This also applies to the situation where there's nothing we can help with for a newly typed #shortcut.

This would avoid the awkward situation where there's a menu that opens with no clickable options as @mgramigna mentioned, while also keeping a more simplified workflow from an interaction standpoint and (eliminates the need for a disabled-when-complete-help-menu).

Some mockups below for more clarification. Added an approach to communicating a list of options when some of them are also disabled (not hoverable = no style changes when hovered). Looks like there's some pros/cons to consider so I'm curious to hear others thoughts.

Screen Shot 2019-09-04 at 2 59 13 PM

Dtphelan1:
Hey Edwin, thanks for the input! There are two other cases that we've been anticipating: 

1) A user adds a lot of incomplete elements to the editor by inserting a template or a snippet (n.b. we don't really support snippets now, but we could imagine doing so). This means we can't assume that people are _necessarily_ confused with what to do; some options in the menu might just be their preferred workflow for completing incomplete shortcuts. 

2) A user wants to edit fields on a complete shortcut and goes back to the shortcut to accomplish that. Since this would allow people to open the GUI on a complete shortcut, we didn't want to abandon showing a menu in a subset of complete-shortcut cases, since for some complete shortcuts (e.g. @condition) users can already return and select a new value. 

For those reasons, I do think that have some menu display even when the menu is complete will support more flexible interactions from users. Let me know if that makes sense - at least I think that's how we've been thinking about it until this point. 

As for the mockups, I think that helps frame the conversation - importantly, I think it raises some more fine-grained questions about interaction! What does hovering over an element look like? Can I arrow down into a disabled element? If I can, what happens when a user hits enter when hovering that element? If I can't, what about when a user mouses over and selects? These more detailed questions are why I punted on the disabled menu items for now in favor of this short term solution.

sgober:
Rebased and pending travis...

edwinychoi:
@Dtphelan1 Thanks for the description! I think that makes sense, I'll revisit the sketches with those additional 2 cases in mind. Current implementation looks good.

PR #762, opened 2019-09-03
support sticky in IE

KeeyanGhoreshi:
This task is actually to fix the bug that the popup for the medications in the calendar don't show up sometimes.  The root cause is that IE doesn't support `position: sticky` and the html element that is used to anchor the popup can move off screen, so the popup does too.  To fix this issue I implemented a simple version of a sticky component for the medication items on the calendar.  It behaves almost identically to the css version of sticky.  It is a bit slower than the css version and has an issue where the table and the values will flicker out of view for a frame.  This can be a little jarring but is small enough that it can be ignored.  The table itself, or at least the header, already flickers when dragging through dates, but these changes for some reason cause it to affect the medication items as well.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [ ] Code is commented

**Tests**

- [ ] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/timeline/Timeline.scss
+++ b/src/timeline/Timeline.scss
@@ -110,9 +110,9 @@ $weekend: rgba(250, 246, 225, 0.5);
     }
 
     .rct-item-content {
-      position: sticky;
+      position: static;
       position: -webkit-sticky;
-      left: 0px;
+
       overflow: hidden;
       display: inline-block;
       border-radius: 2px;
--- a/src/timeline/Timeline.scss
+++ b/src/timeline/Timeline.scss
@@ -112,7 +112,6 @@ $weekend: rgba(250, 246, 225, 0.5);
     .rct-item-content {
       position: sticky;
       position: -webkit-sticky;
-
       overflow: hidden;
       display: inline-block;
       border-radius: 2px;
--- a/src/timeline/Timeline.scss
+++ b/src/timeline/Timeline.scss
@@ -113,6 +113,7 @@ $weekend: rgba(250, 246, 225, 0.5);
       position: sticky;
       position: -webkit-sticky;
       overflow: hidden;
+      max-height:30px;
       display: inline-block;
       border-radius: 2px;
       left: 0;
--- a/src/timeline/Timeline.scss
+++ b/src/timeline/Timeline.scss
@@ -110,7 +110,7 @@ $weekend: rgba(250, 246, 225, 0.5);
     }
 
     .rct-item-content {
-      position: static;
+      position: sticky;
       position: -webkit-sticky;
 
       overflow: hidden;
--- a/src/timeline/TimelineEventsVisualizer.jsx
+++ b/src/timeline/TimelineEventsVisualizer.jsx
@@ -67,24 +67,29 @@ class TimelineEventsVisualizer extends Visualizer {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        if (prevState.interval !== this.state.interval && !supportsSticky) {
-            this.initMedicationItemsPosition(this.state.visibleTimeStart);
+        if (!supportsSticky) {
+            if (prevState.interval !== this.state.interval) {
+                this.initMedicationItemsPosition(this.state.visibleTimeStart);
+            }
+            if (prevState.visibleTimeStart !== this.state.visibleTimeStart) {
+                this.updateMedicationItemsPosition(this.state.visibleTimeStart);
+            }
+            if (prevState.items.length !== this.state.items.length) {
+                this.initMedicationItemsPosition(this.state.visibleTimeStart);
+            }
         }
-        if (prevState.visibleTimeStart !== this.state.visibleTimeStart && !supportsSticky) {
-            this.updateMedicationItemsPosition(this.state.visibleTimeStart);
-        }
-        if (prevState.items.length !== this.state.items.length) {
-            this.initMedicationItemsPosition(this.state.visibleTimeStart);
-        }
+
 
     }
 
     handleLoad = () => {
-        this.calWidth = document.querySelector('[class="react-calendar-timeline"]').getBoundingClientRect().width;
-
         if (!supportsSticky) {
-            document.querySelector(`[class="minimap-container fitted-panel"]`).onscroll = this.scrollCheck;
-            this.initMedicationItemsPosition(this.state.visibleTimeStart);
+            const calendar = document.querySelector('[class="react-calendar-timeline"]');
+            if (calendar !== null && calendar !== undefined) {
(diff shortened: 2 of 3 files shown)


Dtphelan1:
[DP1] - Got some linting errors that need resolving.

src/timeline/Timeline.scss
```diff
@@ -112,7 +112,7 @@ $weekend: rgba(250, 246, 225, 0.5);
     .rct-item-content {
       position: sticky;
       position: -webkit-sticky;
-      left: 0px;
+
```
Dtphelan1:
```suggestion
```

src/timeline/TimelineEventsVisualizer.jsx
```diff
@@ -20,13 +20,12 @@ class TimelineEventsVisualizer extends Visualizer {
         const { conditionSection, tdpSearchSuggestions, highlightedSearchSuggestion, isWide } = this.props;
         const items = this.createItems(conditionSection, tdpSearchSuggestions, highlightedSearchSuggestion);
         const groups = this.createGroupsForItems(this.getMaxGroup(items));
-
+        this.stickySupport = true;
```
Dtphelan1: [DP1] Do we want to have this use some of the utilities defined in Slate for checking what Browser we're using (in `src/lib/slate/constants/environment.js`)? On the one hand that would allow us to actually use regular sticky on other browsers; on the other hand, I don't know we can reasonably expect `window.navigator` to be defined when we launch the app via Smart-on-FHIR. @rdingwell Do you  know anything about how/if the Smart-on-FHIR launcher would scramble any information stored in `window.navigator`?

KeeyanGhoreshi: The sticky support is acquiesced from the CSS computed style, since unsupported "position" defaults to "static".  So this solution lets us use regular sticky on other browsers anyway.  Though if we can detect browsers that will probably be a more comprehensive, less hacky solution.

Dtphelan1: Ehhh I think the way we do this now is fine. I retract this request.

src/timeline/TimelineEventsVisualizer.jsx
```diff
@@ -274,6 +316,23 @@ class TimelineEventsVisualizer extends Visualizer {
         });
     }
 
+    updateMedicationItems() {
```
Dtphelan1: [DP1] This should be an arrow function.

src/timeline/TimelineEventsVisualizer.jsx
```diff
@@ -60,6 +59,27 @@ class TimelineEventsVisualizer extends Visualizer {
         };
     };
 
+    componentDidMount() {
+        window.addEventListener('load', this.handleLoad);
+    }
+
+    componentDidUpdate() {
+        if (!this.stickySupport) {
+            this.updateMedicationItems();
```
Dtphelan1: [DP2] We should use lodash to throttle this somewhat. The performace is super choppy on my IE instance, and I anticipate it'll be marginally better with a throttled call here (by something like 300ms ish?)

src/timeline/TimelineEventsVisualizer.jsx
Dtphelan1: lol sorry this seecond comment was made before seeing your thoughtful response.  My bad!

noranda: What about using a util? Something like:
```
// utils/supportsSticky
const supportsSticky = (() => {
  const node = document.createElement('div');
  node.style.position = 'sticky';
  return node.style.position === 'sticky';
})()

export default supportsSticky;```

src/timeline/TimelineEventsVisualizer.jsx
KeeyanGhoreshi: throttling it will make the text jump, it will only move once every 300ms.  Is that what we want?

Dtphelan1:
I know it’s a not ideal... IMO, compared to the choppiness I think text-jumping is better than text-twitching. Open to opinions though, I’m not too set in my opinion. I’ll send a Gif tomorrow to show you what I’m seeing.   
  

  
  

  
  
>

noranda:
[NB1] There is inconsistency between IE and Chrome with how the timeline now behaves. In Chrome, it displays a 1 year timeline showing the correct labels:
Screen Shot 2019-09-13 at 9 04 33 AM
In IE, on initial reload it displays a 1 year timeline with missing labels:
Screen Shot 2019-09-13 at 9 03 23 AM
In IE, after hovering over an element in the timeline will it then switch to a 3 year timeline showing the correct labels:
Screen Shot 2019-09-13 at 9 04 13 AM

This branch also needs to be rebased.

KeeyanGhoreshi:
I can't seem to recreate the bug.  I'm using IE11 in a VM, I navigate to /pilot1 and wait for the page to load and then scroll down to the timeline and the labels are all where they should be on the 1 year display.  Hovering over the elements works fine as well.  Am I missing something?

noranda:
@KeeyanGhoreshi I was using BrowserStack where I saw this. Perhaps someone on an actual Windows machine can test it?

KeeyanGhoreshi:
The latest commit fixes quite a few bugs too numerous to name, but overall it should work in almost every normal working case.  Though it doesn't address Noranda's bug, since I still can't seem to reproduce it, everything worked fine on the demo machine in the lab, which is a windows machine.

noranda:
@KeeyanGhoreshi Can you rebase this and fix the failed builds?

noranda:
[NB2] The app breaks on the `/ccp` endpoint.

Screen Shot 2019-09-19 at 10 57 05 AM

KeeyanGhoreshi:
Recent commit should deal with NB2

noranda:
[NB1] and [NB2] are resolved, however, there is a failing test now and this needs to be rebased again.

KeeyanGhoreshi:
the rebase should deal with the broken test

noranda:
Looks good now!

Dtphelan1:
The sticky behavior works as expected... however now that all the bugs have been ironed out of this new approach I'm not convinced that this is more performant than the original approach. I'd like to talk a little bit about this in the lab when you get in, but if it's not bringing a lot else to the table then maybe we want to consider.... dare I say.... reverting back to before our big brain play.

Dtphelan1:
On second test, this looks wonderful. Great work Keeyan!

Dtphelan1:
Once this is rebased again we'll merge.

Dtphelan1:
This appears to work after the rebase, so I'm good to merge.

PR #765, opened 2019-09-04
Ansible deploy script

mgramigna:
Covers Jiras 2162 and 2205.

This PR is an Ansible playbook that would automate deployment to an Apache virtual host on the target machine. The use case for now is the fluxnotes.org machine, targeting a host that serves files to some subdomain (likely pilot.fluxnotes.org).

I will provide corresponding documentation for this privately because I'm not sure if it should be on GitHub or not.

**Important**: This PR is more for visual review than anything, as well as seeing if the docs are understandable enough such that someone else could follow along if need be. There are no actual testing steps to be done here.

### DIFF
--- a/deploy.yml
+++ b/deploy.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: dev
+- hosts: pilot
   remote_user: "{{ username }}"
   become_method: sudo
   vars:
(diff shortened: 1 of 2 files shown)


danlee1025:
Looks great!  The documentation is very detailed and looks easy to follow.

PR #766, opened 2019-09-04
initial implementation of click on shortcuts triggering components and shifting focus.

Dtphelan1:
Addresses 2171. Now clicking on a shortcut moves the cursor into an appropriate position.

### DIFF
--- a/src/notes/StructuredFieldPlugin.jsx
+++ b/src/notes/StructuredFieldPlugin.jsx
@@ -605,33 +605,40 @@ function StructuredFieldPlugin(opts) {
             // Need to handle two cases - where keys are arrays and where keys are just strings
             if (_.isArray(keyForShortcut)) {
                 for (const key of keyForShortcut) {
+                    // Add two keys: The key of the node (an Inline) and its children (Text nodes)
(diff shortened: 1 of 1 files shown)


danlee1025:
Oops latest commit failed to compile.

danlee1025:
Looks good.

[DL1]  Only thing I really notice is that it is not super consistent in opening up the completion component, especially when you're switching between 2 incomplete shortcuts.  

For example with `#disease status #status #reason`, switching between #status and #reason doesn't also open the completion component.  Not sure if it's something we want to fix in this PR or another JIRA.

Dtphelan1:
No I think that's a fair comment. I can take a look at it later :)

Dtphelan1:
Rebased this

Dtphelan1:
I'm going to file the DL1 as under work to be done on the completion component, your original intuition @danlee1025 . While it is a weirdness that gets exposed more now, it's in a very different part of the code. I'm going to probably look at it next, but it shouldn't hold this up IMO.

danlee1025:
Ok I'm going to do a quick re-review and approve this.

sgober:
reviewing this now...

PR #782, opened 2019-09-27
[CCP] Fixes side effect header bug when side effect no longer exists

noranda:
This PR satisfies [JIRA 2116](https://jira.mitre.org/browse/ASCODCP-2116) which fixes a bug where the side effect header name would disappear when switching to a filter that removes all side effects of that type. For example, selecting the `erectile dysfunction` side effect filter and then selecting the `gender` demographic similar patients filter would remove the side effect filter from the list and render an empty string in its place.

To fix this, `sideEffects` and `selectedSideEffects` were moved from the component level state into redux and then checked on `UPDATE_PATIENT_OUTCOMES`. If after patient outcomes were updated the selected side effect filter no longer exists, it now reverts it back to `Most Common`.

As part of the refactor, moved `gatherSideEffects` to its own util to prevent unnecessary function calls on every render and to use in the reducer.

Tests were also fixed. Due to the mutating behavior of `sort()`, the `similarPatientTreatmentsData` array in the tests would result in an inconsistent order. The array was duplicated on initialization to prevent this.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


dehall:
[DH1] I stumbled upon this completely by accident but there's a new bug with the checkboxes. Check the top-level option for either the "demographic" or "pathology" section, collapse the section, then re-expand it and uncheck the top-level option. The options become de-selected based on the number of patients returned, but the checkboxes still display as checked. If you click on them once in this state it doesn't appear to do anything but is actually enabling them, then if you click it again it disables and everything is back to normal. If this is the only way to trigger weirdness, it's not a big deal, but I'm concerned there may be other related bugs. I'm not able to reproduce this on fluxnotes.org

KeeyanGhoreshi:
DH1 exists on master as well.

dehall:
So this looks good but I have to wonder, would it be a significant effort to display 0% if the selected side effect no longer exists in the cohort? The current approach feels a little jarring when you change something and then suddenly the side effects change because the previously selected side effect isn't present anymore.

noranda:
@dehall I had initially thought to implement this as you suggest, showing 0% down the table with the selected side effect, but decided against it because the list of side effects is dynamically generated based on the data. If I included the selected side effect that no longer exists, why not then show all possible side effects even though they also don't exist? In order to avoid that rabbit hole, I decided to reset the list anytime the side effect no longer appeared in the data. I agree the sudden change could be a bit distracting, however it is consistent with how the rest of the data in the table also updates to show the current data given the selections.

I rebased this branch and cleaned up some code.

PR #784, opened 2019-09-27
Update parentless shortcuts

jafeltra:

PR for updating a shortcut's parents (ASCODCP-2026).

Basic example for this it to add #nausea, then go back to the front of the line and add #toxicity, and then go back before that to add @condition. As you add parents, shortcuts should become complete. Once there is a condition set, values should appear in the TDP.

You should also be able to reset the parent of a shortcut (change from GIST to Fracture) and children shortcuts should update (ONCOHIST will change based on condition). You should also be able to delete parents without deleting the children and the children will render as incomplete. You should also be able to select all and delete any group of shortcuts. Deleting should update TDP back to its original status. Adding a local context (#toxicity) should not update any related children that are in a different local context (in a different paragraph). Adding a global context should update any shortcuts after it, regardless what paragraph they are in, but no shortcuts before it.

There can be a lot of edge cases with this, so test the shortcuts that you can think of! I think the best way forward with bugs is to report them on the PR and we will decide if they need to be fixed. If they do, we can make subtasks to track them.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code


mgramigna:
[MG1] We should probably remove this console error for no parent contexts. Or perhaps we could change it to a `console.warn` for now?

image

jafeltra:
For MG1, I'd be fine either way. They are technically true statements, so maybe a console.warn makes sense until we have the ability to see why they're incomplete in the editor. Thoughts?

mgramigna:
`console.warn` makes sense to me

jafeltra:
Just rebased after Sarah's PR got merged. Also changed to `console.warn` to address MG1.

mgramigna:
[MG2] Oncohist doesn't update when switching condition

Dtphelan1:
Comments in person as a first pass. Continuing to debug now

jafeltra:
Pushed two changes based on conversations in the lab.

1. Changed disease attribution keyword to illness to avoid conflicts with "disease status" temporarily
2. Added another message to display instead of completion menu with expand option if the shortcut has children but no parent. Previously, the "already complete" message was displayed if a shortcut is complete. Now that a shortcut can be added without it's parent, it can be incomplete but still have been expanded. This is a quick fix to avoid continually expanding shortcuts that are already expanded and the approach should be more thought out in the other Jira for indicating a missing parent (ASCODCP-2102)

karenzheung:
Not really a bug but more of a question/observation
type #stop medication and expand it, and select a medication, Ibuprofen
then on the next line type @active medication but leave it incomplete
Then go back and delete ibuprofen. When you click on @active medication the completion portal doesn't show ibuprofen.
Is this intended?

Dtphelan1:
Karen, I don't think that the behavior you're describing is intended (I'm leaning towards flagging it as a bug @jafeltra, but probably not a priority bug at this time).

jafeltra:
Good catch @karenzheung! I think that happens because the list options gets populated on initialize of the shortcut. When you delete the #stop medication, it doesn't recalculate the list of options for the second @active medication, so it never gets the ibuprofen back into the list.

I feel like this is probably edge-case enough that we can leave it as is for now, as long as you're okay with leaving it too. I can make a Jira to address it later.

karenzheung:
I'm fine with making it a jira. This occurs with @active medication and @active medications

mgramigna:
[MG2] @stage does not work. 

#staging #t1 #n1 #m1 @stage causes a crash. I can help look into this if need be 
image

mgramigna:
Update on MG2, only crashes if you do not have a condition. This is because it calls `getAttributeValue` of the parent context, but staging won't have an attribute value without GIST. Definitely a unique since @stage seems to be the only shortcut that computes its value based on other shortcuts in the editor specifically by getting the attribute value of a parent. Fine to log it as a bug since it's very specific behavior. Might be worth some investigation though who knows /shrug

mgramigna:
[MG3] *non-essential* more for usability but leaving a note here for posteriority. Now that we can insert triggers without any portals, we should probably allow the enter key to also autocomplete a shortcut (right now it is just a space). We have a jira for adding commas and semicolons as autocomplete characters, should probably just add enter too when there is no portal. Again, non-essential.

mgramigna:
[MG4] #nlptoxicity does not complete after the parsing goes through.

image

jafeltra:
Fixed MG2 and MG4.

MG2- @stage wasn't working for a few reasons. It was trying to get a parent context that didn't exist (this was an issue on shortcut entry overhaul too). Added a guard for that. Also once there was a parent, it would disappear because the staging value wasn't set without any T, N, or M values (this has been an issue on master for forever). Added a guard to only return a stage value if it is not an empty string, which leaves the editor with the @stage incomplete text until a stage value is set. Also, needed to get @stage to recalculate if you added a T, N or M value after you had set the parent context (#staging) and inserted (@stage). In order to do that, added an `isSiblingContext` to T, N, and M in order to tell those shortcuts to check for dependent siblings and rerender when they are added.

MG4- #nlptoxicity wasn't being updated to mark that it was incomplete after the NLP engine returned because we weren't updating the parent of the new children shortcuts that were returned from the engine (this was an issue on shortcut entry overhaul). Added the call to update parent context in the insertStructuredFieldAtRange function, similar to regular insertStructuredField.

Thanks to @daniellee1025  and @mgramigna for help with these.

Dtphelan1:
You guys have all the fun once I leave... 😢
  

  
  

  
  
>

mgramigna:
Changes look good. Definitely rando edge cases that we all have contributed to over months that we are uncovering on this PR, though the reputation of your PR remains un-tainted I promise

PR #790, opened 2019-10-10
Improve queries in Smart on FHIR

jafeltra:

Addresses ASCODCP-2193

- Limit the resource types automatically fetched to only ones that include patient or subject
- Get a list of resources that are referenced by the fetched resources
- Fetch any additional resources that were referenced and not yet fetched

This can be tested by observing that fewer resources are returned when loading `/synthea`, but all the data displayed should be the same.

## Submitter:

**Running the Application**

**Task**

**Documentation**

**Code Quality**

- [x] Code is commented

**Tests**

- [x] Existing tests passed

## Reviewers:

- Try to break the code

### DIFF
--- a/src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
+++ b/src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
@@ -37,6 +37,11 @@ const sampleObservationSearchData = {
     entry: hardCodedFHIRPatient.entry.filter(e => e['resource']['resourceType'] === 'Observation')
 };
 
+const sampleReferenceResult = {
+    resourceType: 'Bundle',
+    entry: hardCodedFHIRPatient.entry.filter(e => e['resource']['resourceType'] === 'Condition')
+}
+
 
 describe('SMART on FHIR data source', function() {
     const originalWindowFHIR = window.FHIR;
@@ -83,7 +88,9 @@ describe('SMART on FHIR data source', function() {
           .get('/fhir/Patient?_id=1078857')
           .reply(200, samplePatientSearchData)
           .get('/fhir/Observation?patient=1078857')
-          .reply(200, sampleObservationSearchData);
+          .reply(200, sampleObservationSearchData)
+          .get('/fhir/Encounter/6a8bc97e-3ba1-4fb5-a478-4b5bd888c793') // reference retrieval
+          .reply(200, sampleReferenceResult);
           // in this case it doesn't need to fetch the metadata since the resourceTypes are manually specified
 
         const dataSource = new McodeV09SmartOnFhirDataSource({ resourceTypes: ['Patient', 'Observation'] });
--- a/src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
+++ b/src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
@@ -37,10 +37,7 @@ const sampleObservationSearchData = {
     entry: hardCodedFHIRPatient.entry.filter(e => e['resource']['resourceType'] === 'Observation')
 };
 
-const sampleReferenceResult = {
-    resourceType: 'Bundle',
-    entry: hardCodedFHIRPatient.entry.filter(e => e['resource']['resourceType'] === 'Condition')
-}
+const sampleReferenceResult = hardCodedFHIRPatient.entry.filter(e => e['resource']['resourceType'] === 'Encounter');
 
 
 describe('SMART on FHIR data source', function() {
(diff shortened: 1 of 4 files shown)


src/dataaccess/McodeV09SmartOnFhirDataSource.js
```diff
@@ -108,7 +142,29 @@ class McodeV09SmartOnFhirDataSource extends IDataSource {
...
+                    .then((resources) => {
+                        // After the first pass of queries are resolved, fetch any additional resources that were
+                        // referenced that have not yet been fetched.
+
+                        // Filter out any searchset bundles that don't have any entries
+                        nonEmptyResources = nonEmptyResources.concat(resources.filter(res => res.entry));
+                        this.referencedEntries.forEach((key) => {
+                            const referenceArray = key.split('/');
+                            const resourceType = referenceArray[0];
+                            const resourceId = referenceArray[1];
```
mgramigna: I don't think references are guaranteed to always be of the format `<resource type/<id>`. For example, they could be full URLs, like `http://<fhir server>/<resource type>/<id>`. In this case, doing a split on `'/'` would not work they way it's intended to attempt to parse out the resource type and the id from a reference.

Not sure of the best way to approach being smarter about this. We could use a regex, or always just look at the things prior to and after the last slash? Not sure.

src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
```diff
@@ -37,6 +37,11 @@ const sampleObservationSearchData = {
     entry: hardCodedFHIRPatient.entry.filter(e => e['resource']['resourceType'] === 'Observation')
 };
 
+const sampleReferenceResult = {
+    resourceType: 'Bundle',
+    entry: hardCodedFHIRPatient.entry.filter(e => e['resource']['resourceType'] === 'Condition')
+}
+
```
Dtphelan1: [DP1] Since this is the expected response of a id-specific request, it should be the resource itself and not a bundle with the resource as an entry. For example, look at the response of `http://moonshot-dev.mitre.org:4002/baseDstu2/Patient/smart-1032702`

src/dataaccess/McodeV09SmartOnFhirDataSource.js
```diff
@@ -68,13 +74,33 @@ class McodeV09SmartOnFhirDataSource extends IDataSource {
             .then(response => this._handleResponseBundle(response.data, prevBundle));
     }
 
+    _handleResponseEntry(entry) {
+        if (entry.resourceType && entry.resourceType === 'Bundle') return entry;
```
Dtphelan1: [DP2] Can we add a comment here about what this function is doing? Looks like it's mapping straight fhir resources into bundle-like objects - is that right?

src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
```diff
@@ -83,7 +88,9 @@ describe('SMART on FHIR data source', function() {
           .get('/fhir/Patient?_id=1078857')
           .reply(200, samplePatientSearchData)
           .get('/fhir/Observation?patient=1078857')
-          .reply(200, sampleObservationSearchData);
+          .reply(200, sampleObservationSearchData)
+          .get('/fhir/Encounter/6a8bc97e-3ba1-4fb5-a478-4b5bd888c793') // reference retrieval
+          .reply(200, sampleReferenceResult);
           // in this case it doesn't need to fetch the metadata since the resourceTypes are manually specified
 
         const dataSource = new McodeV09SmartOnFhirDataSource({ resourceTypes: ['Patient', 'Observation'] });
```
Dtphelan1: [DP3] Does this need "Encounter" to be added?

src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
jafeltra: You're definitely correct. I will update the format.

src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
jafeltra: "Encounter" does not _need_ to be added here because the Encounter is being fetched due to the fact that it is a reference on one of the entries that does match those types. When we fetch references, we do not check against the specified types (if there are any). I will add a Jira to explore replacing these specified types on props in config.js with just a set list of supported types that is used everywhere.

jafeltra:
MG1, DP1, and DP2 should all be addressed.

I believe there is nothing that needs to be changed for DP3 because we do not need to add the types of references to the list of types we pass on the datasource props. I did make a Jira to explore if we can get rid of this list that specifies types in favor of a list that is used everywhere to limit the types we query for (https://jira.mitre.org/browse/ASCODCP-2282).

src/__test__/backend/dataaccess/McodeV09SmartOnFhirDataSource.test.js
Dtphelan1: 👍

